PR #31, opened 2021-02-02
draft: try to package as Flatpak

likern:


likern:
This is my **best** try to package key-mapper as Flatpak application.
It can be run using the following commands:
```bash
flatpak-builder --force-clean --user --install flatpak-build/ flatpak/io.github.sezanzeb.KeyMapper.yml
flatpak run io.github.sezanzeb.KeyMapper
```

Unfortunately, this **doesn't work** because application **requires enabled systemd service**.
And I can't find any information how to provide these permissions. As I understand that's not possible now. 

https://github.com/flatpak/flatpak/issues/2787 I think might be related.

This PR is only for future references, as I spend whole day reading documentation and FreeDesktop specifications.
I tried to apply *all the best and latest practices* I could find until that error - That's why these strange names "io.github....". Hope this mught be usefull for reference.

flatpak/io.github.sezanzeb.KeyMapper.metainfo.xml
```diff
@@ -0,0 +1,55 @@
...
+      <image>https://raw.githubusercontent.com/sezanzeb/key-mapper/main/readme/screenshot.png</image>
+    </screenshot>
+    <screenshot>
+      <image>https://raw.githubusercontent.com/sezanzeb/key-mapper/main/readme/screenshot_2.png</image>
+    </screenshot>
+  </screenshots>
+
+  <url type="homepage">https://github.com/sezanzeb/key-mapper</url>
+  <url type="contact">proxima@hip70890b.de</url>
+  <update_contact>proxima@hip70890b.de</update_contact>
```
sezanzeb, line 53: changed my domain to sezanzeb.de. email proxima@sezanzeb.de

flatpak/key-mapper.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+PYTHONPATH="/app/usr/lib/python3.8/site-packages/" python3.8 /app/usr/bin/key-mapper-gtk
```
sezanzeb, line 3: does flatpak deliver all python versions, or is it possible that somebody would have 3.7 instead of 3.8?

sezanzeb:
Please add `/app/usr/share/key-mapper` to `candidates` in data.py

flatpak/io.github.sezanzeb.KeyMapper.yml
```diff
@@ -0,0 +1,23 @@
+app-id: io.github.sezanzeb.KeyMapper
+runtime: org.gnome.Platform
+runtime-version: '3.38'
+sdk: org.gnome.Sdk
+command: /app/bin/key-mapper.sh
+
+finish-args:
```
sezanzeb, line 7: `  - --device=all` is needed to see anything in the gui, but that won't make all devices visible unfortunately

sezanzeb: But it still fails at injecting stuff

```
Process Process-1:
  File "/usr/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/usr/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/app/usr/lib/python3.8/site-packages/keymapper/dev/injector.py", line 439, in _start_injecting
    path, uinput.capabilities(verbose=True)
  File "/app/lib/python3.8/site-packages/evdev/uinput.py", line 224, in capabilities
    raise UInputError('input device not opened - cannot read capabilities')
evdev.uinput.UInputError: input device not opened - cannot read capabilities
```

which is most likely caused by missing permissions on `/dev/uinput`

flatpak/key-mapper.sh
likern: These are isolated containers. If there's python3.8, as I understand it should be always there. It's what is provided as a standard Gnome environment https://docs.flatpak.org/en/latest/available-runtimes.html.

So these are not host files, by files inside virtual machine / container.

flatpak/io.github.sezanzeb.KeyMapper.yml
likern: > No access to any device nodes

I think this is the problem https://docs.flatpak.org/en/latest/sandbox-permissions.html

likern: @sezanzeb Have you tried this option?
> --device=all

`--device=DEVICE`
Expose a device to the application. This updates the [Context] group in the metadata. DEVICE must be one of: dri, kvm, shm, all. This option can be used multiple times.

likern: According to https://docs.flatpak.org/en/latest/flatpak-command-reference.html this should work
> All device nodes in /dev, but not /dev/shm (which is separately specified). Available since 0.6.6.

likern: Sorry, misread your message. If `device=all` doesn't work - could you open bug in Flatpak describing original intent of application.

sezanzeb: https://github.com/flatpak/flatpak/issues/4137

sezanzeb:
see https://github.com/sezanzeb/key-mapper/issues/21#issuecomment-908633246

Aqua1ung:
There's [a new GUI tool](https://flathub.org/apps/details/com.github.tchx84.Flatseal) that might [help with permissions](https://www.techrepublic.com/article/manage-flatpak-permissions-flatseal/) for apps packaged as flatpak. Maybe you guys could take another stab at it?

bayazidbh:
Hi, for anyone working on a Flatpak release, the [latest Flatpak pre-release](https://github.com/flatpak/flatpak/releases/tag/1.15.6) seems to have added support for [`--device=input permission`](https://github.com/flatpak/flatpak/pull/5481). Though the [full USB device access permission](https://github.com/flatpak/xdg-desktop-portal/pull/559) is still WIP.

PR #45, opened 2021-02-19
Add systemctl enable for PIP install

pbaumard:
Thank you for this great tool!

After my PIP installation, I just noticed the `sudo systemctl enable key-mapper` was missing in installation steps to get the configuration back after reboot (same steps as https://github.com/sezanzeb/key-mapper/blob/main/DEBIAN/postinst).


sezanzeb:
thanks!

Issue #40, opened 2021-02-13
Packing via debian/rules


sezanzeb:
To satisfy FTBFSIASW (https://ftp-master.debian.org/REJECT-FAQ.html). Apparently it should be possible to create .deb packages using `dpkg-buildpackage -us -uc`. scripts/build.sh would not be allowed

It is some sort of makefile, and I know nothing about those. I also seems quite troublesome to get it running like that. It would be nice if somebody who has done debian compliant deb packaging in the past could help out here.

sezanzeb:
examples for other python packages:
- https://packages.ubuntu.com/hirsute/soundconverter
- https://packages.ubuntu.com/hirsute/lutris
- https://packages.ubuntu.com/hirsute/menulibre
- https://packages.ubuntu.com/hirsute/quodlibet

`debian.tar.xz` file on the right

sezanzeb:
There are 3 people until now who said they would want this in official repos. I'd love to get it into the debian repo

skitt:
Feel free to assign this to me, I’ve started working on it.

sezanzeb:
Awesome! Thank you

So ideally we will first get a debian compliant build running, and then with version 1.0.0 request to get it into the repo (see https://mentors.debian.net/sponsors/)

skitt:
OK, that works for me. There won’t be any need to find a sponsor, I’m an uploading DD so I can submit packages directly to the archive (or, more accurately, the “NEW queue” for review by the FTP team who are responsible for the archive overall).

Blaimi:
summary of #82: The development of the debian-package for building via debian/rules has moved to [debian salsa](https://salsa.debian.org/python-team/packages/key-mapper). There is also a ITP-Report for debian: [#986918](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=986918)

sezanzeb:
1.0.0 released, I'm fine with putting that into the debian repo

skitt:
Excellent, I’ll take care of that tomorrow.

sezanzeb:
really amazing.

I committed the version bump after I made the release tag this time. Don't know if that makes any difference for you, just to let you know.

skitt:
It’s in the NEW queue (waiting for review by the FTP team): https://ftp-master.debian.org/new/key-mapper_1.0.0-1.html

sezanzeb:
[...]

PR #82, opened 2021-04-09
Draft: first try to debianize

Blaimi:
Closes: #40


Blaimi:
@skitt: you mentioned, you have already started with packaging (https://github.com/sezanzeb/key-mapper/issues/40#issuecomment-796573301). Would you mind to review this commit and merge your stuff if necessary? I think we also need an ITP-report.

I don't know, if the (build-)dependencies are complete or partially unnecessary, I did not (yet) test the package on a fresh installation or other than ubuntu 20.04.

Lintian warns about missing man-pages but that's a different topic. We should also consider switching to quilt-format to be more flexible when porting to different distribution-version (e.g. debian/ubuntu, backports, etc.) or package-formats (I'd love to see this also as rpm for fedora).

debian/rules
```diff
@@ -0,0 +1,19 @@
...
+	rm -f debian/key-mapper.service
+
+override_dh_auto_install:
+	dh_auto_install
+	# remove the systemd-service in /usr/lib/systemd and replace it with dh-systemd
+	rm -rf debian/key-mapper/usr/lib/systemd/
+	cp data/key-mapper.service debian/key-mapper.service
+	# move udev-rule from /etc to /lib for packaging
+	mkdir -p debian/key-mapper/lib
+	mv debian/key-mapper/etc/udev debian/key-mapper/lib/udev
```
sezanzeb, line 16: should this be changed in setup.py?

Blaimi: I don't know whats best practice for pip-packages. Debian packages use `/lib/udev` and lintian complains if it detects a udev-rule in a different directory.

sezanzeb: yeah lets change that in setup.py instead https://unix.stackexchange.com/a/333714

Blaimi: Sounds good. Do you make the commit? I'll rebase this PR afterwards.

skitt:
> @skitt: you mentioned, you have already started with packaging ([#40 (comment)](https://github.com/sezanzeb/key-mapper/issues/40#issuecomment-796573301)). Would you mind to review this commit and merge your stuff if necessary? I think we also need an ITP-report.

Hah, I was working on the packaging in the usual Debian way, _i.e._ in a separate, Debian-hosted repo.

> I don't know, if the (build-)dependencies are complete or partially unnecessary, I did not (yet) test the package on a fresh installation or other than ubuntu 20.04.
> 
> Lintian warns about missing man-pages but that's a different topic. We should also consider switching to quilt-format to be more flexible when porting to different distribution-version (e.g. debian/ubuntu, backports, etc.) or package-formats (I'd love to see this also as rpm for fedora).

debian/changelog
```diff
@@ -0,0 +1,5 @@
+key-mapper (0.8.2) UNRELEASED; urgency=low
```
skitt, line 1: This shouldn’t be a “native” package, so:

```suggestion
key-mapper (0.8.2-1) UNRELEASED; urgency=low
```

debian/control
```diff
@@ -0,0 +1,29 @@
+Source: key-mapper
+Section: misc
+Priority: optional
+Maintainer: Sezanzeb <proxima@sezanzeb.de>
+Standards-Version: 4.5.0
```
skitt, line 5: The current version is 4.5.1:

```suggestion
Standards-Version: 4.5.1
```

debian/control
```diff
@@ -0,0 +1,29 @@
...
+Build-Depends:
+ debhelper-compat (= 12),
```
skitt, line 7: For a Debian package you should target the currently-recommended version:

```suggestion
 debhelper-compat (= 13),
```

If this packaging is supposed to target multiple releases, this might not be appropriate. It’s not an issue for Debian 10, `debhelper` 13 is available as a backport; it might be a problem for some releases of Ubuntu.

debian/control
```diff
@@ -0,0 +1,29 @@
...
+ dh-python,
+ libpython3-dev, libdbus-1-dev,
+ python3,
```
skitt, line 10: This should be

```suggestion
 python3-all,
```

debian/control
```diff
@@ -0,0 +1,29 @@
+Source: key-mapper
+Section: misc
+Priority: optional
+Maintainer: Sezanzeb <proxima@sezanzeb.de>
+Standards-Version: 4.5.0
+Build-Depends:
+ debhelper-compat (= 12),
+ dh-python,
+ libpython3-dev, libdbus-1-dev,
```
skitt, line 9: This shouldn’t be necessary.

```suggestion
```

debian/control
```diff
@@ -0,0 +1,29 @@
...
+ python3,
+ python3-evdev,
+ python3-pydbus,
+ python3-gi,
```
skitt, line 13: I think you’ll also need

```suggestion
 python3-gi,
 python3-setuptools,
```

debian/control
```diff
@@ -0,0 +1,29 @@
...
+Vcs-Git: https://github.com/sezanzeb/key-mapper
+Vcs-Browser: https://github.com/sezanzeb/key-mapper
+
+Package: key-mapper
+Architecture: all
+Depends: python3-evdev,
+ python3-pydbus,
```
skitt, line 20: These two get filled in by `${python3:Depends}:

```suggestion
Depends:
```

debian/control
```diff
@@ -0,0 +1,29 @@
...
+ python3-gi,
+ ${misc:Depends},
+ ${python3:Depends},
+Replaces: python3-key-mapper
+Description: tool to change the mapping of your input device buttons
+ An easy to use tool to change the mapping of your input device buttons.
+ .
+ Supports mice, keyboards, gamepads, X11, Wayland,
+ combined buttons and programmable macros.
```
skitt, line 29: This is what I came up with:

```suggestion
Description: Input device button mapping tool
 key-mapper allows users to map buttons on all input devices
 (keyboards, mice, gamepads...) in X11 and Wayland. It also supports
 combined buttons and programmable macros.
 .
 key-mapper includes a UI to configure button mappings, per device,
 and configuration to automatically apply button mappings at boot and
 on device connection.
```

debian/copyright
```diff
@@ -0,0 +1,17 @@
...
+License: GPL-3+
+
+Files: debian/*
+Copyright: 2020-2021 sezanzeb
+  2021 Matthias Blümel <debian@blaimi.de>
+License: GPL-3+
+
+License: GPL-3+
+ On Debian systems the full text of the GNU GPL v2 can be found
+ in the `/usr/share/common-licenses/GPL-3' file.
```
skitt, line 17: You still need the license grant here:

```suggestion
License: GPL-3+
 This package is free software; you can redistribute it and/or modify
 it under the terms of the GNU General Public License as published by
 the Free Software Foundation; either version 3 of the License, or
 (at your option) any later version.
 .
 This package is distributed in the hope that it will be useful,
 but WITHOUT ANY WARRANTY; without even the implied warranty of
 MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 GNU General Public License for more details.
 .
 You should have received a copy of the GNU General Public License
 along with this program. If not, see <http://www.gnu.org/licenses/>
 .
 On Debian systems, the complete text of the GNU General
 Public License version 3 can be found in "/usr/share/common-licenses/GPL-3".
```

debian/rules
skitt, line 16: You can avoid all this by adding symlinks permanently in `debian`:

```
key-mapper.service -> ../data/key-mapper.service
key-mapper.udev -> ../data/key-mapper.rules
```

This ensures that the maintainer scripts get the added snippets required to correctly handle services.

I also added a patch in `debian/patches` to remove the two lines in `setup.py` which install the files.

debian/source/format
```diff
@@ -0,0 +1 @@
+3.0 (native)
```
skitt, line 1: Please make this

```suggestion
3.0 (quilt)
```

skitt (changes requested):
I also patched `setup.py` to remove the commit hash update from `logger.py` (otherwise the build tools complain of untracked changes to the source code).

debian/changelog
Blaimi: Yes. Like I said, we should switch to quilt.

What's your opinion on to the hosting? github or salsa? And which tool is your preferred one? `git-buildpackage`?

Can you provide the patch?

debian/control
Blaimi: I ran it on ubuntu 20.04 and got a lintian warning: `key-mapper source: newer-standards-version 4.5.1 (current is 4.5.0)` so I downgraded.

debian/control
Blaimi: Ubuntu 20.04 still has 12. I don't think we should depend on backport-packages if we can avoid it.

debian/rules
Blaimi: Sounds good. @sezanzeb wants to change the udev-path in setup.py anyway. I'd write a patch for the service when this gets migrated to quilt.

debian/source/format
Blaimi: see comment above. We should also add a watch-file.

sezanzeb:
> I also patched `setup.py` to remove the commit hash update from `logger.py` (otherwise the build tools complain of untracked changes to the source code).

I could write the hash into a gitignored file instead. I'll try to look into that on the weekend

debian/rules
Blaimi: Ah, didn't realize that udev also has a debhelper. nice

debian/changelog
skitt: I pushed my packaging tree to https://salsa.debian.org/python-team/packages/key-mapper if you want to take a look. I prefer `gbp`, yes, and would follow the Python team guidelines for this project.

debian/control
skitt: The purpose of the field is to track the version of the standards which it adheres to; your packaging is compliant with 4.5.1, so it should be declared as such, even if older versions of lintian don’t know about it.

debian/control
skitt: That makes sense.

debian/rules
skitt: Yes, `dh_installudev`.

debian/source/format
skitt: https://salsa.debian.org/python-team/packages/key-mapper/-/blob/master/debian/watch works for me ;-).

sezanzeb:
> I could write the hash into a gitignored file instead. I'll try to look into that on the weekend (including changing the udev path)

done, pushed to main

sezanzeb:
had to change it to `/usr/lib` for arch/manjaro. `/lib` seems to be a symlink to `/usr/lib` in ubuntu as well though so I hope that also works for you

Blaimi:
> had to change it to `/usr/lib` for arch/manjaro. `/lib` 

That's fine. I'll use `dh_installudev` anyway ;)

sezanzeb:
I think key-mapper pretty much has everything it would need for a stable release. But it's too late for bullseye right? https://release.debian.org/bullseye/freeze_policy.html (https://github.com/sezanzeb/key-mapper/issues/83#issuecomment-816748037)

skitt:
New packages are accepted at any time into Debian experimental or unstable; it’s just testing that’s frozen currently. The packaging is ready on Salsa IMO, I’m just poking at the tests to see if I can enable them before uploading to the NEW queue. @Blaimi, would you like me to add you as a co-maintainer?

Blaimi:
Once it is accepted in unstable, we can get it into bullseye-backports or even buster-backports.

@skitt: I'd skip the „co“-prefix ;).

skitt:
> @skitt: I'd skip the „co“-prefix ;).

I’m not sure what you mean...

sezanzeb:
great. Lets wait for at most one more month to see if any new bug pops up, I'll release 1.0.0 soon then.

Blaimi:
> > @skitt: I'd skip the „co“-prefix ;).
> 
> I’m not sure what you mean...

I think I misunderstood what you meant. If you mean this repository, you have to ask @sezanzeb

I finally found the time to review the contents of the salsa-repo. It looks much better than my draft ;).

I'd create a ubuntu-ppa for the package but I think this requires to decrease the debhelper-compat to 12 for building on focal, The packages could also be built on unstable and then copied, I don't know how fine this works with python-packages.

I'm also closing this PR for now since the development moved to salsa.

Ah, and would you mind to enable the [pipeline](https://salsa.debian.org/salsa-ci-team/pipeline)? Or is this unusual for python-packages?

skitt:
> > > @skitt: I'd skip the „co“-prefix ;).
> > 
> > I’m not sure what you mean...
> 
> I think I misunderstood what you meant. If you mean this repository, you have to ask @sezanzeb

I was asking if you wanted to be listed as a co-maintainer on the Debian package (along with me and the Python team).

> I finally found the time to review the contents of the salsa-repo. It looks much better than my draft ;).

Thanks!

> I'd create a ubuntu-ppa for the package but I think this requires to decrease the debhelper-compat to 12 for building on focal, The packages could also be built on unstable and then copied, I don't know how fine this works with python-packages.

What we could do is get the Salsa repo sorted, get the package into Debian (which acts as a quality check to some extent), and then copy the `debian` directory into this repo if @sezanzeb is OK with that, and downgrade the dependencies so that the package can be built in a focal PPA.

> Ah, and would you mind to enable the [pipeline](https://salsa.debian.org/salsa-ci-team/pipeline)? Or is this unusual for python-packages?

Yup, I’ll do that.

sezanzeb:
> then copy the debian directory into this repo

yes, sounds good. the deb packaging that is currently in this repo can absolutely be replaced with your solution. The readme will have to be updated as well then.

PR #106, opened 2021-04-27
Add GObject dependencies to setup.py

skitt:
key-mapper needs gi, GLib, and Gtk at runtime; this ensures they are available.

Signed-off-by: Stephen Kitt <steve@sk2.org>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -81,7 +81,8 @@ setup(
     install_requires=[
         'setuptools',
         'evdev',
-        'pydbus'
+        'pydbus',
+        'pygobject',
     ],
     cmdclass={
         'install': Install,


sezanzeb:
```
Requirement already satisfied: setuptools in /usr/lib/python3.9/site-packages (from key-mapper==0.8.1) (54.1.1)
Requirement already satisfied: evdev in /usr/lib/python3.9/site-packages (from key-mapper==0.8.1) (1.4.0)
Requirement already satisfied: pydbus in /usr/lib/python3.9/site-packages (from key-mapper==0.8.1) (0.6.0)
Requirement already satisfied: pygobject in /usr/lib/python3.9/site-packages (from key-mapper==0.8.1) (3.38.0)
ERROR: Could not find a version that satisfies the requirement gi.repository.GLib>=2.0 (from key-mapper)
ERROR: No matching distribution found for gi.repository.GLib>=2.0
```

via `sudo pip install .`

sezanzeb:
Lutris seems to use `PyGObject` only (https://github.com/lutris/lutris/blob/master/setup.py)

skitt:
My bad, it does indeed fail with plain `pip` (which probably explains why the dependencies are commented out in [Solaar’s `setup.py`](https://github.com/pwr-Solaar/Solaar/blob/master/setup.py)). I’ve reduced the PR to `pygobject` only.

Issue #120, opened 2021-06-06
Tap and Hold binding for same button?


TRPB:
This may already be possible but I couldn't quite piece it together from the documentation.

- On tap of a joystick button (e.g. press and release within 0.5s) pass the key through to the game as normal (so it gets triggered on release, not press)
- On hold (e.g. button is held down longer than 0.5s) send a configured mapping

I tried the following:

```
set("pressed", 1).h(w(500).set("pressed",0)).ifeq("pressed", 1, k(x), k(y))
```

The problem is, it always waits 500ms then presses Y. If `h()` exited immediately (and halted any current waits) on key release this should work. 

Thanks for the great tool!

edit: I've found the source of the issue but I'm not really sure how to fix it:

In  `macros.py` line 290 there's this:

```py
 if isinstance(macro, _Macro):
            # repeat the macro forever while the key is held down
            async def task(handler):
                while self.is_holding():
                    # run the child macro completely to avoid
                    # not-releasing any key
                    await macro.run(handler)
```

When the key is released, it runs whatever is inside `h()` so that always gets run. The problem is, I don't want it to be  run. I need something like the following:

```py
    async def run(self, handler, endOnWait=True):
        """Run the macro.
        Parameters
        ----------
        handler : function
            Will receive int type, code and value for an event to write
        """
        if self.running:
            logger.error('Tried to run already running macro "%s"', self.code)
            return
        
        self.keystroke_sleep_ms = self.mapping.get('macros.keystroke_sleep_ms')

        self.running = True
        for task in self.tasks:
            # one could call tasks the compiled macros. it's lambda functions
            # that receive the handler as an argument, so that they know
            # where to send the event to.
            coroutine = task(handler)
            if asyncio.iscoroutine(coroutine):
                if (endonWait && coroutine == wait) break;
                else await coroutine

        # done
        self.running = False

```

sezanzeb:
try `set("pressed", 1).w(500).h(set("pressed",0)).ifeq("pressed", 1, k(x), k(y))`

hsantanna:
You can call hold inside hold. That way the expression will continue to execute only if the key is still pressed.

This will do X **with no delay** if fast press and release (<150ms) the button and do X with a bit of delay if release between 150ms and 500ms, but will do Y on release if hold key for >500ms:
`set(pressed, 1).w(150).h(w(350).h(set(pressed, 0).w(50))).ifeq(pressed, 1, k(x), k(y) )`

[...]

PR #121, opened 2021-06-07
Halt waits in h() when key is released - enables different actions depending on key press length

TRPB:
I created  an issue #120 for this but I was able to implement it myself.

This enables the creation of macros which do different things depending on how long the button has been pressed:

# Functionality

## Long press / short press

```
set("pressed", 1).h(w(500).set("pressed",0)).ifeq("pressed", 1, k(x), k(y))
```

If the button is released within 500ms it prints `x` if it's released after 500ms it prints `y`. 

## Tap / hold

```
set("pressed", 1).h(w(100).set("pressed",0).h(y)).ifeq("pressed", 1, k(x), k(y))
```

On tap (button released within 100ms) prints `x` after 100ms it acts like `y` is being held down. 

# Implementation

The main change is in the `run` method:

```python
        for task in self.tasks:
            # one could call tasks the compiled macros. it's lambda functions
            # that receive the handler as an argument, so that they know
            # where to send the event to.
            coroutine = task(handler)

            if (self.is_holding() == False and coroutine.__name__ == "sleep"):
                break

            if asyncio.iscoroutine(coroutine):
                await coroutine
                if (self.is_holding() == False and coroutine.__name__ == "sleep"):
                    break
```

If the button has been released and the next task is `wait`   then it stops running any further tasks. E.g. if the button is released during the wait of `w(500).k(y)`  then `k(y)` will never get executed. This is done twice because the key could be released before the loop runs or during the `await`.

The second change is the sleep function:

```py
        async def sleep(_):
            """ Wait in intervals of 10ms so the wait can be ended early if the key is released """
            for i in range(int(sleeptime/10)):
                await asyncio.sleep(0.01)
                if (self.is_holding() == False):
                    break

```

Instead of `await asyncio.sleep(sleeptime/1000)` it chunks the waits into 10ms intervals. Without this in the example above, tapping and releasing the key would make the `x` appear only after the wait timer had finished (e.g. 500ms). This reduces the input lag to 10ms max.

### DIFF
--- a/keymapper/injection/macros.py
+++ b/keymapper/injection/macros.py
@@ -226,8 +226,14 @@ class _Macro:
             # that receive the handler as an argument, so that they know
             # where to send the event to.
             coroutine = task(handler)
+
+            if (self.is_holding() == False and coroutine.__name__ == "sleep"):
+                break
+
             if asyncio.iscoroutine(coroutine):
                 await coroutine
+                if (self.is_holding() == False and coroutine.__name__ == "sleep"):
+                    break
 
         # done
         self.running = False
@@ -447,10 +453,14 @@ class _Macro:
                 f'a number, but got "{sleeptime}"'
             ) from error
 
-        sleeptime /= 1000
-
+       
         async def sleep(_):
-            await asyncio.sleep(sleeptime)
+            """ Wait in intervals of 10ms so the wait can be ended early if the key is released """
+            for i in range(int(sleeptime/10)):
+                await asyncio.sleep(0.01)
+                if (self.is_holding() == False):
+                    break
+                
 
         self.tasks.append(sleep)
 


sezanzeb:
Thanks for contributing! I'll review your code shortly

This comment is actually deprecated:

> run the child macro completely to avoid not-releasing any key

It was valid when the `def keycode` function was adding multiple tasks, one for key-down, for wait, and for key-up. Stopping the macro early could have potentially stopped it after the key-down, which would result in a key that is held down forever. It was changed to be a single task at some point.

So I think your change works

TRPB:
Playing with this further, the first if statement in `run` can be removed as I added that before I modified the sleep function.

I'm not sure how other people are using macros but this change could potentially break some macros. If you had `k(x).w(500).k(x)` with this change, x would only appear twice if the button was held down.

Perhaps it needs an argument? `w(500, True)` would exit if the key is released whereas `w(500, False)` or `w(500)` would keep running the chain after the key had been released?

sezanzeb:
that reminds me of https://github.com/sezanzeb/key-mapper/issues/109

if there was more space for editing macros it would be totally fine to add new functions with longer names, or to use meaningful flags like `hold(..., breakable)`.

Not sure if I would find it intuitive that a `w` function has a special meaning for `h`, since I wouldn't expect that if it was treated like a lambda function in python.

sezanzeb:
Any updates? There are quite a lot of ways this could be solved in the future and things to consider, so lets discuss it a bit first before making a decision on it

sezanzeb:
this is covered with if_tap now

PR #127, opened 2021-06-15
Extended docs by foot pedals and keyboard scroll

sarimak:
Thanks for the key-mapper! It greatly increased the value of Kinesis Savant Elite 3-pedal and allowed me to map the scrolling to unused keyboard keys which both are great helpers in fighting with RSI.

### DIFF
--- a/readme/examples.md
+++ b/readme/examples.md
@@ -119,31 +119,6 @@ while the joystick turns the view (depending on the game).
 Tested with the XBOX 360 Gamepad. On Ubuntu, gamepads worked better in
 Wayland than with X11.
 
-## Combinations spanning multiple devices
-
-For regular combinations on only single devices it is not required to
-configure macros. See [readme/usage.md](usage.md#combinations).
-
-**Keyboard** `space` `set(foo, 1).h(space).set(foo, 0)`
-
-**Mouse** `middle` `if_eq($foo, 1, h(a), h(BTN_MIDDLE))`
-
-Apply both presets. If you press space on your keyboard, it will write a
-space exactly like it used to. If you hold down space and press the middle
-button of your mouse, it will write "a" instead. If you just press the
-middle button of your mouse it behaves like a regular middle mouse button.
-
-**Explanation**
-
-`h(space)` makes your key work exactly like if it was mapped to "space".
-It will inject a key-down event if you press it, does nothing as long you
-hold your key down, and injects a key-up event after releasing.
-`set(foo, 1).set(foo, 0)` sets "foo" to 1 and then sets "foo" to 0.
-`set` and `if_eq` work on shared memory, so all injections will see your
-variables. Combine both to get a key that works like a normal key, but that also
-works as a modifier for other keys of other devices. `if_eq($foo, 1, ..., ...)`
-runs the first param if foo is "bar", or the second one if foo is not "bar".
-
 ## Sequence of Keys with Modifiers
 
 Alt+TAB, Enter, Alt+TAB:
--- a/readme/examples.md
+++ b/readme/examples.md
@@ -1,0 +1,180 @@
+# Examples
+
+Examples for particular devices and/or use cases:
+
+## Event Names
+
+- Alphanumeric `a` to `z` and `0` to `9`
+- Modifiers `Alt_L` `Control_L` `Control_R` `Shift_L` `Shift_R`
+- Mouse buttons `BTN_LEFT` `BTN_RIGHT` `BTN_MIDDLE` `BTN_SIDE` ...
+- Multimedia keys `KEY_NEXTSONG` `KEY_PLAYPAUSE` `XF86AudioMicMute` ...
+- Mouse scroll `wheel(down, 1)` `wheel(up, 1)`
+- Mouse move `mouse(left, 1)` `mouse(right, 1)` `mouse(up, 1)` `mouse(down, 1)`
+
+## Macros
+
+- `k(BTN_LEFT)` a single mouse-click
+- `k(1).k(2)` 1, 2
+- `r(3, k(a).w(500))` a, a, a with 500ms pause
+- `m(Control_L, k(a).k(x))` CTRL + a, CTRL + x
+- `k(1).h(k(2)).k(3)` writes 1 2 2 ... 2 2 3 while the key is pressed
+- `e(EV_REL, REL_X, 10)` moves the mouse cursor 10px to the right
+- `mouse(right, 4)` which keeps moving the mouse while pressed.
+  Made out of `h(e(...))` internally
+- `wheel(down, 1)` keeps scrolling down while held
+- `set(foo, 1)` set ["foo"](https://en.wikipedia.org/wiki/Metasyntactic_variable) to 1
+- `ifeq(foo, 1, k(x), k(y))` if "foo" is 1, write x, otherwise y
+- `h()` does nothing as long as your key is held down
+- `h(a)` holds down "a" as long as the key is pressed, just like a regular non-macro mapping
+- `if_tap(k(a), k(b))` writes a if the key is tapped, otherwise b
+- `if_tap(k(a), k(b), 1000)` writes a if the key is released within a second, otherwise b
+- `if_single(k(a), k(b))` writes b if another key is pressed, or a if the key is released
+  and no other key was pressed in the meantime.
+
+## Combinations Spanning Multiple Devices
+
+For regular combinations on only single devices it is not required to
+configure macros. See [readme/usage.md](usage.md#combinations).
+
+**Keyboard** `space` `set(foo, 1).h(space).set(foo, 0)`
+
+**Mouse** `middle` `if_eq($foo, 1, h(a), h(BTN_MIDDLE))`
+
+Apply both presets. If you press space on your keyboard, it will write a
+space exactly like it used to. If you hold down space and press the middle
+button of your mouse, it will write "a" instead. If you just press the
+middle button of your mouse it behaves like a regular middle mouse button.
+
+**Explanation**
+
+`h(space)` makes your key work exactly like if it was mapped to "space".
+It will inject a key-down event if you press it, does nothing as long you
+hold your key down, and injects a key-up event after releasing.
+`set(foo, 1).set(foo, 0)` sets "foo" to 1 and then sets "foo" to 0.
+`set` and `if_eq` work on shared memory, so all injections will see your
+variables. Combine both to get a key that works like a normal key, but that also
+works as a modifier for other keys of other devices. `ifeq($foo, 1, ..., ...)`
+runs the first param if foo is 1, or the second one if foo is not 1.
+
+
+## Scroll and Click on a Keyboard
+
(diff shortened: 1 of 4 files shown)


sezanzeb:
Thanks! Mouse and wheel movements are actually macros and examples are here: https://github.com/sezanzeb/key-mapper/blob/main/readme/macros.md

Do you think that this is hard to find?

Maybe the docs should mention that the number is the speed

readme/usage.md
```diff
@@ -99,6 +109,13 @@ while the joystick turns the view (depending on the game).
 Tested with the XBOX 360 Gamepad. On Ubuntu, gamepads worked better in
 Wayland than with X11 for me.
 
+## Foot Pedals
+
+Foot pedal clicks can be translated into mouse scroll events even if the
+device didn't support that. Tested on Kinesis Savant Elite 2 (3-pedal
+version) on Wayland/Ubuntu. Scroll events generated while holding the
+pedal linearly auto-repeat as they should.
```
sezanzeb, line 117: But that applies to all kinds of events (multimedia, keys, numbers, etc.), not just scrolling and not just foot-pedals. This is basically the whole purpose of key-mapper, so it should be written somewhere else. Maybe a generalized sentence can be placed at the top of usage.md, like "In theory any click, joystick or wheel event of any device can be mapped to any of the available key names, even if the device didn't support that", if you feel like it wasn't clear already.

> Scroll events generated while holding the pedal linearly auto-repeat as they should.

Basically https://github.com/sezanzeb/key-mapper/blob/main/readme/macros.md already says "keeps scrolling down while held"

readme/usage.md
```diff
@@ -85,7 +85,17 @@ Examples:
...
 - Mouse buttons `BTN_LEFT` `BTN_RIGHT` `BTN_MIDDLE` `BTN_SIDE` ...
-- Multimedia keys `KEY_NEXTSONG` `KEY_PLAYPAUSE` ...
+- Mouse scroll `wheel(down, 1)` `wheel(up, 1)`
+- Mouse move `mouse(left, 1)` `mouse(right, 1)` `mouse(up, 1)` `mouse(down, 1)`
+- Multimedia keys `KEY_NEXTSONG` `KEY_PLAYPAUSE` `XF86AudioMicMute` ...
+
+## Keyboard/Mouse cooperation
+
+Left click mapped to a keyboard key can be held while moving the cursor by
+a real mouse/trackball, resulting in proper drag'n'drop.
```
sezanzeb, line 95: This isn't something that key-mapper implemented, it's just how the environment uses the provided events mapped by the user. If a BTN_LEFT can be mapped then it would be expected that it behaves just like a regular mouse button

also this is kinda an example, it could be also said about other combinations of injected events like *shift on device 1* + *a on device 2*. What is it about the *btn_left on device 1* + *movement on device 2* combination that is so special that it should be mentioned?

readme/usage.md
```diff
@@ -85,7 +85,17 @@ Examples:
...
+
+Mouse scroll up/down mapped to keyboard keys can add scroll support to cheap
+3-button mice and trackballs without a scroll ring.
```
sezanzeb, line 98: Or gamepads, foot pedals, any device. By showing the possibility to map those kinds of events with key-mapper in the example section it should be clear that it can be added to any device. Not sure if this addition adds value

sezanzeb (commented):
Not so sure about this, sorry :/, it feels a bit redundant and examply

readme/usage.md
sarimak: Aha, so the idea behind the current documentation is to keep it brief and normalized/deduplicated, do I understand it correctly? My initial impression from the docs was they are too terse and written from the developer point of view which doesn't that much care about highlighting practical usage of such great tool for someone who just discovered it and is trying to judge if the tool fits into his needs -- or is not aware of the full potential of such tool.

Would it make sense to you to split the docs into technical description of commandline options, syntax of key names and macros etc. (existing docs) -- and into (newly created) howto guide which would showcase concrete examples of use with different types of devices? Perhaps it could also improve the discoverability of this tool via web search engines.

sezanzeb: ok, lets add another doc file for it.

As far as usage.md and macros.md goes yeah, I think I tried to have that a bit generalized. There is a little bit of example stuff in a few places, but nothing that gives very concrete examples like "on this trackball model if you map that you can have this and that and it even works as foo".

Underknowledge:
Just scrolling through but I vote for a +1 for a excessive `examples.md`
I spend about 2 hours to find the right combination for an alt+tab enter alt+tab 
`"m(Alt_L, k(tab)).w(250).k(KP_Enter).k(KP_Enter).k(key_UP).w(150).m(Alt_L, k(tab))"`

sezanzeb:
@sarimak unfortunatelly hasn't gotten back to this since https://github.com/sezanzeb/key-mapper/pull/127#discussion_r652554809

sarimak:
Oops, I forgot to follow up, thanks for reminding me. Added to my todo, will have a look when time permits.

sarimak:
Just for sure: Could you please summarize your idea how should I modify the PR so it would fit into the documentation scheme? Create a new file `examples.md` and just move the content there? Should I move any existing content (like the mention about D-pad) there as well?

sezanzeb:
> The D-Pad can be mapped to W, A, S, D for example (...)

yes, lets move that to examples.md

There should be a link in the readme to examples.md, just like the other docs

sezanzeb:
there is also an example on the bottom of https://github.com/sezanzeb/key-mapper/blob/main/readme/macros.md that should probably be moved to examples.md

sarimak:
I tried to incorporate all the comments into the PR and squashed it. Please have a look if I forgot/mis-interpreted anything.

@ Underknowledge: I added your example as-is. Could you please check if the double Enter is really OK? Could you also please explain your use case why this particular key combination is useful enough for you to have it mapped to a key (so I can expand the paragraph by it).

readme/examples.md
```diff
@@ -0,0 +1,154 @@
...
+In contrast to a real scroll wheel, holding a key which has mouse wheel event
+mapped produces linear auto-repeat, without any acceleration. Using a PageDown
+key for fast scrolling requires only a small adjustment of the right hand
+position.
+
+## Scroll on a 3-Button Mouse
+
+Cheap 3-button mouse without a scroll wheel can scroll using the middle button:
+
+- Button MIDDLE: wheel(down, 1)
```
sezanzeb, line 63:
```suggestion
- Button MIDDLE: `wheel(down, 1)`
```

readme/examples.md
```diff
@@ -0,0 +1,154 @@
...
+- Button SIDE: BTN_RIGHT
+
+## Scroll on Foot Pedals
+
+While Kinesis Savant Elite 2 foot pedals can be programmed to emit key press or
+mouse click events, they cannot emit scroll events themselves. Using the pedals
+for scrolling while standing at a standing desk is possible thanks to remapping:
+
+- Button LEFT: wheel(up, 1)
+- Button RIGHT: wheel(down, 1)
```
sezanzeb, line 82:
```suggestion
- Button LEFT: `wheel(up, 1)`
- Button RIGHT: `wheel(down, 1)`
```

readme/examples.md
```diff
@@ -0,0 +1,154 @@
...
+Wayland than with X11.
+
+## Combinations spanning multiple devices
+
+For regular combinations on only single devices it is not required to
+configure macros. See [readme/usage.md](usage.md#combinations).
+
+**Keyboard** `space` `set(foo, bar).h(space).set(foo, 0)`
+
+**Mouse** `middle` `ifeq(foo, bar, h(a), h(BTN_MIDDLE))`
```
sezanzeb, line 103:
```suggestion
**Keyboard** `space` `set(foo, 1).h(space).set(foo, 0)`

**Mouse** `middle` `if_eq($foo, 1, h(a), h(BTN_MIDDLE))`
```

if_eq supports the new `$variable` syntax, ifeq should not be mentioned in the docs anymore. I couldn't migrate ifeq to support that feature without breaking the way it previously worked.

readme/examples.md
```diff
@@ -0,0 +1,154 @@
...
+**Explanation**
+
+`h(space)` makes your key work exactly like if it was mapped to "space".
+It will inject a key-down event if you press it, does nothing as long you
+hold your key down, and injects a key-up event after releasing.
+`set(foo, 1).set(foo, 0)` sets "foo" to 1 and then sets "foo" to 0.
+`set` and `ifeq` work on shared memory, so all injections will see your
+variables. Combine both to get a key that works like a normal key, but that also
+works as a modifier for other keys of other devices. `ifeq(foo, bar, ..., ...)`
+runs the first param if foo is "bar", or the second one if foo is not "bar".
```
sezanzeb, line 119:
```suggestion
`h(space)` makes your key work exactly like if it was mapped to "space".
It will inject a key-down event if you press it, does nothing as long you
hold your key down, and injects a key-up event after releasing.
`set(foo, 1).set(foo, 0)` sets "foo" to 1 and then sets "foo" to 0.
`set` and `if_eq` work on shared memory, so all injections will see your
variables. Combine both to get a key that works like a normal key, but that also
works as a modifier for other keys of other devices. `if_eq($foo, 1, ..., ...)`
runs the first param if foo is "bar", or the second one if foo is not "bar".
```

readme/macros.md
```diff
@@ -1,6 +1,6 @@
 # Macros
```
sezanzeb, line 1: I made a lot of changes to this file. Just pull in the latest from main and remove the example on the bottom.

sezanzeb (changes requested):
Great, thanks a lot!

I just added some `backticks` here and there and left some advice for merging in the latest changes in main.

Underknowledge:
@sarimak 
It's ok, but maybe for an example not optimal. Might want to get rid of the extra enter keypress. but I guess it's not a massive downer.
I start a molecule testing job via an old-ass apple trackpad with it (tap with 4 fingers to start the test job). 
the extra enter is to have a new line to better see the different runs.

sarimak:
@Underknowledge I removed the extra Enter but kept the example as-is.

@sezanzeb Hope I applied all changes correctly, please have look if it is OK now.

readme/examples.md
```diff
@@ -0,0 +1,180 @@
...
+
+`h(space)` makes your key work exactly like if it was mapped to "space".
+It will inject a key-down event if you press it, does nothing as long you
+hold your key down, and injects a key-up event after releasing.
+`set(foo, 1).set(foo, 0)` sets "foo" to 1 and then sets "foo" to 0.
+`set` and `if_eq` work on shared memory, so all injections will see your
+variables. Combine both to get a key that works like a normal key, but that also
+works as a modifier for other keys of other devices. `if_eq($foo, 1, ..., ...)`
+runs the first param if foo is "bar", or the second one if foo is not "bar".
+
```
sezanzeb, line 146: This section appears twice

```suggestion
```

sezanzeb (changes requested):
I'll merge it after this

PR #129, opened 2021-06-28
Add initial i18n support

jose1711:


bin/key-mapper-gtk
```diff
@@ -24,13 +24,25 @@
...
 gi.require_version('GLib', '2.0')
 from gi.repository import Gtk
 
+APP_NAME = 'key-mapper'
+LOCALE_DIR = os.path.join(get_data_path(), 'lang')
+
+locale.bindtextdomain(APP_NAME, LOCALE_DIR)
+locale.textdomain(APP_NAME)
+
+print('Using locale directory: {}'.format(LOCALE_DIR))
```
sezanzeb, line 44: Can you use the logger module instead of print here? 

`logger.debug('Using locale directory: %s', LOCALE_DIR)`

setup.py
```diff
@@ -55,6 +61,27 @@ def get_packages():
...
+            subprocess.run(['msgfmt', '-o', mo_path, str(po_file)], check=True)
+            mo_files.append(mo_path)
+
+        return mo_files
+
+lang_data = []
+for po_file in glob.glob(PO_FILES):
+    lang = splitext(basename(po_file))[0]
+    lang_data.append(('/usr/share/key-mapper/lang/{}/LC_MESSAGES'.format(lang),
+                      ['{}/LC_MESSAGES/key-mapper.mo'.format(lang)]))
```
sezanzeb, line 83: f-strings

I also prefer not to use hanging indents. Imho this is more readable, especially when things get more complicated than that.

```suggestion
    lang_data.append((
        f'/usr/share/key-mapper/lang/{lang}/LC_MESSAGES',
        [f'{lang}/LC_MESSAGES/key-mapper.mo']
    ))
```

setup.py
```diff
@@ -86,5 +115,6 @@ def get_packages():
     ],
     cmdclass={
         'install': Install,
+        'refresh_langs': RefreshLangs
```
sezanzeb, line 118: Does this require another command to be executed before key-mapper is fully installed? If yes, can this be added to `Install`?

sezanzeb (changes requested):
Thank you so much!

Just some minor styling tweaks and one open question

setup.py
jose1711: Done

setup.py
jose1711: > Does this require another command to be executed before key-mapper is fully installed? If yes, can this be added to `Install` instead?

Yes, it requires user/admin to run `setup.py refresh_langs` when the translations are needed. I am a bit reluctant to put this into `Install` as the current approach gives more flexibility. After making a commit I found https://github.com/oleksis/youtube-dl-gui/blob/1dbc65d5fb7ca45038efaaf48259976eb4319099/setup.py#L267 where they're doing a similar thing. If you're still convinced this belongs to `Install` I will put it there ofc.

sezanzeb: quodlibet also uses `./setup.py build` as a separate command: https://quodlibet.readthedocs.io/en/latest/development/overview.html

lets call the command `build` as well

the new command should be part of
- the installation instructions for pip https://github.com/sezanzeb/key-mapper/blob/main/README.md
- the build script https://github.com/sezanzeb/key-mapper/blob/main/scripts/build.sh
- and the AUR package (I'll do that)

setup.py
```diff
@@ -55,6 +61,29 @@ def get_packages():
...
+            mo_files.append(mo_path)
+
+        return mo_files
+
+lang_data = []
+for po_file in glob.glob(PO_FILES):
+    lang = splitext(basename(po_file))[0]
+    lang_data.append((
+        f'/usr/share/key-mapper/lang/{lang}/LC_MESSAGES',
+        [f'{lang}/LC_MESSAGES/key-mapper.mo']
```
sezanzeb, line 84: can you put all .mo files into a separate directory? `mo/` or something

setup.py
sezanzeb: I think after this and the mo folder this is probably ready to merge, thank you :)

jose1711: >     * the installation instructions for pip https://github.com/sezanzeb/key-mapper/blob/main/README.md

Sadly I've got no idea how can I instruct `pip` to run desired `setup.py` functions. Browsing tutorial did not reveal anything either :disappointed:

sezanzeb: either

```
git clone https://github.com/sezanzeb/key-mapper.git
cd key-mapper
python3 setup.py build
sudo pip3 install .
```

or put it into install after all to make `pip3 install git+https://github.com/sezanzeb/key-mapper.git` possible

setup.py
jose1711: done

setup.py
jose1711: > or put it into install after all to make pip3 install git+https://github.com/sezanzeb/key-mapper.git possible

OK, I give up, will go this route.

PR #167, opened 2021-08-17
Use $HOME environment variable to find user's home.

pjamargh:


sezanzeb:
That makes perfect sense, thanks!

Please write some small test for `test_user.py` *(pull from master to get it, I just extracted it from test_paths.py)*

something like:

```py
    def test_get_home(self):
        os.environ['HOME'] = ...
        self.assertEqual(get_home(...), ...)
```

to get all of that juicy test coverage. For root, normal user and HOME set.

pjamargh:
Will do.

sezanzeb:
```python
➜  ~ pkexec python3
Python 3.9.5 (default, May 11 2021, 08:20:37) 
[GCC 10.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.environ['USER']
'root'
>>> os.environ['HOME']
'/root'
```

in this case home is not the home of the user who called pkexec

pjamargh:
You're right, this doesn't work I expected to work. Without much knowledge of key-mapper structure I think there could be 2 potential kind of solutions:
 1. At start (by the target user) before acquiring privileges via pkexec use environment variables to take note of the user and home directory to be passed as parameters (so there is no complexity to find that out later). Is this possible?
 2. Find out in creative ways (like the get_user). Maybe peeking at /etc/passwd?  On my system there is no entry for my user (super customized corp Linux, not very representative). Anyhow this is best effort, you can always fallback to /home/${USER}

I do not know why I got into this, I think this is only a problem on my current system. I'd say 99% of Linux user will be fine with /home/user.

sezanzeb:
1.

yes, the part of the gui that runs via pkexec can receive messages via a pipe: https://github.com/sezanzeb/key-mapper/blob/main/keymapper/gui/helper.py#L89

But this seems to work as well, so probably no complicated handshake between gui and helper is needed:

```py
➜  ~ sudo useradd test
[sudo] password for mango: 
➜  ~ sudo usermod -d /home/foo test
➜  ~ pkexec python3
Python 3.9.5 (default, May 24 2021, 12:50:35) 
[GCC 11.1.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.path.expanduser('~test')
'/home/foo'
```

2.

> I'd say 99% of Linux user will be fine with /home/user.

probably yes, but if there is a way to look at the actual home dir of a specific user it would still be better than now

sezanzeb:
```py
HOME = os.path.expanduser(f'~{USER}')
```

are you going to push this?

Otherwise I'll make the change myself

pjamargh:
Hey Tobi, will send a different CL, no need to use HOME environment variable as you can take a look at the PWD database.

PR #178, opened 2021-08-31
Use the appropriate user home directory.

pjamargh:
Some issues with the mock library forced me to make some changes.

I also had to tweak test_get_user test as there is no UID 1000 in my system and it was failing because of it. Please run the tests prior to merge, somehow I cannot run them now..:|

### DIFF
--- a/keymapper/user.py
+++ b/keymapper/user.py
@@ -58,8 +58,18 @@ def get_user():
     return user
 
 
+def get_home(user):
+    """Try to find the user's home directory."""
+    try:
+        return pwd.getpwnam(user).pw_dir
+    except:
+        pass
+
+    return '/root' if USER == 'root' else f'/home/{user}'
+
+
 USER = get_user()
 
-HOME = '/root' if USER == 'root' else f'/home/{USER}'
+HOME = get_home(USER)
 
 CONFIG_PATH = os.path.join(HOME, '.config/key-mapper')
--- a/keymapper/user.py
+++ b/keymapper/user.py
@@ -60,13 +60,7 @@ def get_user():
 
 def get_home(user):
     """Try to find the user's home directory."""
-    try:
-        return pwd.getpwnam(user).pw_dir
-    except:
-        pass
-
-    return '/root' if USER == 'root' else f'/home/{user}'
-
+    return pwd.getpwnam(user).pw_dir
 
 USER = get_user()
 
--- a/tests/testcases/test_user.py
+++ b/tests/testcases/test_user.py
@@ -23,8 +23,7 @@ import os
 import unittest
 from unittest import mock
 
-from keymapper.user import get_user
-from keymapper.user import get_home
+from keymapper.user import get_user, get_home
 
 from tests.test import quick_cleanup
 
--- a/tests/testcases/test_user.py
+++ b/tests/testcases/test_user.py
@@ -55,6 +55,7 @@ class TestUser(unittest.TestCase):
             os.environ['PKEXEC_UID'] = '1000'
             self.assertNotEqual(get_user(), 'root')
 
+    def test_get_home(self):
         property_mock = mock.Mock()
         property_mock.configure_mock(pw_dir='/custom/home/foo')
         with mock.patch('pwd.getpwnam', return_value=property_mock):
--- a/tests/testcases/test_user.py
+++ b/tests/testcases/test_user.py
@@ -55,10 +55,6 @@ class TestUser(unittest.TestCase):
             os.environ['PKEXEC_UID'] = '1000'
             self.assertNotEqual(get_user(), 'root')
 
-    def test_get_home(self):
-        with mock.patch('pwd.getpwnam', return_value = None):
-            self.assertEqual(get_home('foo'), '/home/foo')
-
         property_mock = mock.Mock()
         property_mock.configure_mock(pw_dir='/custom/home/foo')
         with mock.patch('pwd.getpwnam', return_value=property_mock):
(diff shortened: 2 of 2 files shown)


pjamargh:
Never mind about the tests. They one changed (test_user) is running fine on my environment.

keymapper/user.py
```diff
@@ -58,8 +58,18 @@ def get_user():
     return user
 
 
+def get_home(user):
+    """Try to find the user's home directory."""
+    try:
+        return pwd.getpwnam(user).pw_dir
+    except:
```
sezanzeb, line 65: is there reason to believe that this might fail? i.e. are the except and the fallback return statement below needed?

tests/testcases/test_user.py
```diff
@@ -24,6 +24,7 @@
 from unittest import mock
 
 from keymapper.user import get_user
+from keymapper.user import get_home
```
sezanzeb, line 27:
```suggestion
from keymapper.user import get_user, get_home
```

tests/testcases/test_user.py
```diff
@@ -43,7 +44,9 @@ def test_get_user(self):
         with mock.patch('os.getlogin', lambda: 'root'):
             self.assertEqual(get_user(), 'root')
 
-        with mock.patch('os.getlogin', lambda: _raise(OSError())):
+        property_mock = mock.Mock()
+        property_mock.configure_mock(pw_name='quix')
+        with mock.patch('os.getlogin', lambda: _raise(OSError())), mock.patch('pwd.getpwuid', return_value = property_mock):
```
sezanzeb, line 49:
```suggestion
        with mock.patch('os.getlogin', lambda: _raise(OSError())), mock.patch('pwd.getpwuid', return_value=property_mock):
```

tests/testcases/test_user.py
```diff
@@ -52,3 +55,13 @@ def test_get_user(self):
...
             os.environ['PKEXEC_UID'] = '1000'
             self.assertNotEqual(get_user(), 'root')
+
+    def test_get_home(self):
+        with mock.patch('pwd.getpwnam', return_value = None):
+            self.assertEqual(get_home('foo'), '/home/foo')
+
+        property_mock = mock.Mock()
+        property_mock.configure_mock(pw_dir='/custom/home/foo')
+        with mock.patch('pwd.getpwnam', return_value = property_mock):
```
sezanzeb, line 65:
```suggestion
        with mock.patch('pwd.getpwnam', return_value=property_mock):
```

pep8

sezanzeb (changes requested):
just some minor stuff

keymapper/user.py
pjamargh: Not that I'm aware of but I'm afraid it might and decided to keep the old behavior (just didn't want to break your code). I'm okay removing it.

sezanzeb: yeah lets remove it then

pjamargh:
BTW, there is something called black and isort that are great to enforce formatting. Maybe you can consider adding them to your project. There is a huge benefit on this being automatic (and enforced). You can see more info and how to add them in this blog post: https://blog.kartones.net/post/four-horsemen-python-apocalypse

sezanzeb:
> black

I just tried the live playground and agree with the style. For me, the most important thing is that it doesn't format stuff like this:

```
foooooooooooooooooooooo(foooooooooooooooooooooo(foooooooooooooooooooooo(a, b, c, d, e), b, c, d, 
                                                foooooooooooooooooooooo(a, b, c, d, e), b,
                                                [1, 2, 3, 4, 5, 6, 7], d, e, b, c, d, e),
                        b, c, d, e)
```

which for some reason is a popular way to indent function calls and such.

Most changes by black seem to be because it uses double-quotes for strings. Would black be something like a pre-commit hook?

pjamargh:
Even if it does indent it that way (until you find the right configuration that allows you to change it) I think the benefits outweigh these issues.

Black and isort are configurable, you might find the appropriate switch also for the quotes. I like them single but I do not care enough to force black to use them.

keymapper/user.py
pjamargh: Done

pjamargh:
I think all is resolved now. I'm not sure why Github still says there are "Changes requested"..

tests/testcases/test_user.py
```diff
@@ -52,3 +54,13 @@ def test_get_user(self):
             del os.environ['SUDO_USER']
             os.environ['PKEXEC_UID'] = '1000'
             self.assertNotEqual(get_user(), 'root')
+
+    def test_get_home(self):
+        with mock.patch('pwd.getpwnam', return_value = None):
+            self.assertEqual(get_home('foo'), '/home/foo')
```
sezanzeb, line 60: Those two lines are obsolete then

sezanzeb: this actually fails. But we don't have a reason to believe it would also fail like that during runtime

pjamargh: Mmm.. I expected this to fail (no more try and you're None is provided) but somehow tests run fine. I think it is probably using the package installed on my computer for the test or something like that.

Yes, this will not fail at runtime (we determined that when we removed the try block + fallback).

I'll remove this test.

tests/testcases/test_user.py
```diff
@@ -55,10 +55,6 @@ def test_get_user(self):
             os.environ['PKEXEC_UID'] = '1000'
             self.assertNotEqual(get_user(), 'root')
 
-    def test_get_home(self):
```
sezanzeb: `def test_get_home(self):`

this line was actually allright. Since you are testing get_home below.

sezanzeb: anyway, I don't care that much. If this is ready I'll hit merge, just let me know

sezanzeb:
thank you

PR #219, opened 2021-11-30
allow autoload command to execute if `config_dir` is provided

tmclane:
I have a Linux machine that does not show any `logged in users` with the `who` command even when I am using the machine.
It is perhaps a specialized use-case but I don't see why you need a logged in user in order to activate the mapping IF a config directory has been provided which contains the relevant configuration for key-mapper.

The machine in question is a JingPad A1.  

```
tmclane@JingOS:/media/tmclane/data/key-mapper$ who
tmclane@JingOS:/media/tmclane/data/key-mapper$ whoami
tmclane
```

### DIFF
--- a/bin/key-mapper-control
+++ b/bin/key-mapper-control
@@ -205,7 +205,7 @@ def main(options):
 
     logger.debug('Call for "%s"', sys.argv)
 
-    if options.command == AUTOLOAD and num_logged_in_users() == 0:
+    if options.command == AUTOLOAD and num_logged_in_users() == 0 and options.config_dir is None:
         # No user logged in, this is probably happening during boot time and got
         # triggered by udev. There is no need to try to inject anything if the
         # service doesn't know where to look for a config file. This avoids a lot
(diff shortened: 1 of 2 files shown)


sezanzeb:
You are absolutely right!

But that won't fix autoloading at login (data/key-mapper-autoload.desktop) or device plugging (data/99-key-mapper.rules) on your system. Any ideas?

It's not critically important to not run key-mapper-control if no user is logged in. It just avoids some confusion when people see the log

sezanzeb:
I have pushed a change to main that uses systemd-analyze, does that work for you?

tmclane:
Let me give your changes a try.

FYI: I found another solution which was to put a script in: `/usr/lib/pm-utils/sleep.d` that calls the `key-mapper-control` command on sleep / wake to disable and reenable it.

Specifically this one: https://github.com/tmclane/jingpad/blob/main/99_jingpad_wakeup

sezanzeb:
I think this can be closed, the change with `systemd-analyze` probably fixed this

Issue #221, opened 2021-11-30
Using Control_L + Alt_L as modifiers requires a waiting time for key to be inserted on Wayland


FrauH0lle:
Thank you so much for creating this @sezanzeb ! I spend hours today to find a replacement for xdotool on Wayland! And this here finally works!

---

Please install the newest version from source to see if the problem has already been solved.

Using commit 6a68b89 

Share some logs please:

1. `key-mapper-control --version`
```
key-mapper 1.2.1  https://github.com/sezanzeb/key-mapper
python-evdev 1.4.0
```
2. which linux distro (ubuntu 20.04, manjaro, etc.)
```
Gentoo
```
3. `echo $XDG_SESSION_TYPE`
```
wayland
```
4. which desktop environment (gnome, plasma, xfce4, etc.)
```
KDE Plasma 5.23.3
```
5. `sudo ls -l /proc/1/exe`
```
lrwxrwxrwx 1 root root 0 30. Nov 14:09 /proc/1/exe -> /lib/systemd/systemd
```
6. paste the affected preset .json file from ~/.config/key-mapper/presets
```
{
    "mapping": {
        "1,29,1+1,56,1+1,8,1": "wait(300).m(ISO_Level3_Shift,k(7))",
        "1,29,1+1,56,1+1,9,1": "wait(300).m(ISO_Level3_Shift,k(8))",
        "1,29,1+1,56,1+1,10,1": "wait(300).m(ISO_Level3_Shift,k(9))",
        "1,29,1+1,56,1+1,27,1": "wait(300).m(ISO_Level3_Shift,k(plus))",
        "1,29,1+1,56,1+1,12,1": "wait(300).m(ISO_Level3_Shift,k(ssharp))",
        "1,29,1+1,56,1+1,11,1": "wait(300).m(ISO_Level3_Shift,k(0))"
    }
}
```

So the idea is to copy the behavior from Windows which allows to use the left Ctrl and Alt key as a replacement for the AltGr key on German keyboards. 
Here:
```
Control_L + Alt_L + 7 -> AltGr + 7 = {
Control_L + Alt_L + 8 -> AltGr + 8 = [
Control_L + Alt_L + 9 -> AltGr + 9 = ]
Control_L + Alt_L + 0 -> AltGr + 0 = }
Control_L + Alt_L + ß -> AltGr + ß = \
Control_L + Alt_L + + -> AltGr + + = ~
```
Testing `Control_L + Alt_L + 7`:

[...]

Issue #198, opened 2021-10-26
Give support to button holding lock


CarlosEkisde:
It would be nice having an option which when switched on made that pressing a particular button would be equal to hold it for the virtual device, and when pressed again, it would be released. It would work for some usecases particularly disabled persons or just for convenience.

sezanzeb:
and

`if_unreleased(a)`

would help

`if_unreleased(a, key_down(a), key_up(a))`

but the naming of if_unreleased might be ambiguous, considering a fictional macro of `if_held`, which might refer to a key being held down on the keyboard.

sezanzeb:
similar: https://github.com/sezanzeb/input-remapper/issues/382#issuecomment-1116319803

sezanzeb:
added `key_down` and `key_up`:

```
if_eq(
    $mode,
    1,
    key_up(KEY_A).set(mode, 0),
    key_down(KEY_A).set(mode, 1)
)
```

PR #228, opened 2021-12-08
Independent uinput

jonasBoss:
here is a proof of concept for https://github.com/sezanzeb/key-mapper/discussions/227
this creates one keyboard uinput device, which emits all the remapped keys and buttons from all input devices.
Apparently we are able to write from multiple processes to the same UInput I don't know why, and I am not sure if it is safe. We probably should still investigate how to run everything in the same process with threading or asyncio

### Tests
an overview to the changes in the tests please edit for more changes or additions

- [x] `TestInjector`
  - [x] `test_adds_ev_key`: removed, the injector no longer controls the capabilities of a uinput
  - [x] `test_gamepad_capabilities`: removed, see above
  - [x] `test_gamepad_purpose_none`: removed checks for `_construct_capabilities`
  - [x] `test_gamepad_purpose_none_2`: removed checks for `_construct_capabilities`
  - [x] `test_store_permutations_for_macros`: patch  Injector with `_copy_capabilities` instead of `_construct_capabilities`
  - [x] `test_capabilities_and_uinput_presence`: no longer test a mapped uinput
- [x] `TestModifyCapabilities`: removed all tests for _construct_capabilities
- [x] `TestMacros.test_handle_plus_syntax`: `handle_plus_syntax` now raises a ValueError for invalid syntax
- [x] `TestGlobalUinputs`: added
- [x] `TestMigrations.test_migrate_mappings`: modified for new migration behavior


keymapper/uinput_device.py
```diff
@@ -0,0 +1,38 @@
...
+
+class UinputDevice:
```
sezanzeb, line 8: what is the benefit of the wrapper over a factory?

keymapper/injection/consumers/consumer.py
```diff
@@ -52,10 +52,10 @@ def is_enabled(self):
...
 
-    def write(self, key):
+    def write(self, key, target="keyboard"):
         """Shorthand to write stuff."""
-        self.context.uinput.write(*key)
-        self.context.uinput.syn()
+        self.context.uinput_devices[target].uinput.write(*key)
+        self.context.uinput_devices[target].uinput.syn()
 
     def forward(self, key):
```
sezanzeb, line 60: I just tried doing the combination modifier fix that involves adding a `k(Shift_L)`, but with `e(EV_KEY, KEY_LEFTSHIFT, 0)`.

Using `e` for that is only possible if the forward function also uses the `uinput_devices["keyboard"].uinput` device

keymapper/uinput_device.py
jonasBoss: I did not know the concept of a factory. My programming knowledge is only self-taught in my free-time. 
Even though I now read some articles about the idea and implementation of a factory-method I am still not sure how it could be useful in this case. 
I think we can decide how we go about creating the uinput object once we know how we want to load them from config files and if a user will be able to manipulate them or create their own.

sezanzeb: > My programming knowledge is only self-taught in my free-time

You appear quite knowledgable actually

sezanzeb: > once we know how we want to load them from config files and if a user will be able to manipulate them or create their own.

the globalUInputs class from that other pr could contain that functionality

sezanzeb: as for loading config files, there could be a folder in ~/.config/key-mapper/uinputs/, containing various json configurations for uinputs.

GlobalUInputs would then cycle through them based on some priority and then figures out if the event should go into that uinput.

keymapper/injection/global_uinputs.py
```diff
@@ -0,0 +1,85 @@
...
+        
+        Wrong events might be injected if the group mappings are wrong
+        """
+        # TODO check for event value especially for EV_ABS
+        try:
+            return event[0] in self.capabilities().keys() and event[1] in self.capabilities()[event[0]]
+        except evdev.uinput.UInputError:
+            logger.debug("uinput for %s is not available", self.name)
+            self.device = self._find_device()
+            return event[0] in self.capabilities().keys() and event[1] in self.capabilities()[event[0]]
```
jonasBoss, line 45: @sezanzeb I get an exception here: `evdev.uinput.UInputError: input device not opened - cannot read capabilities`.

But only when running key-mapper-service directly as a user. As a systemd service everything works as expected.
Is there a way we can set the correct permissions? 

When creating a uinput from the python console it has some extra permissions:
```
$ getfacl /dev/input/event21
getfacl: Entferne führende '/' von absoluten Pfadnamen
# file: dev/input/event21
# owner: root
# group: input
user::rw-
user:jonas:rw-
group::rw-
mask::rw-
other::---
```
compared to the permissions for key-mapper devices:
```
$ getfacl /dev/input/event20
getfacl: Entferne führende '/' von absoluten Pfadnamen
# file: dev/input/event20
# owner: root
# group: input
user::rw-
group::rw-
other::---
```

keymapper/injection/global_uinputs.py
```diff
@@ -0,0 +1,85 @@
...
+        super().__init__( *args, **kwargs)
+    
+    def can_emit(self, event):
+        """ceck it a event can be emitted by the uinput
+        
+        Wrong events might be injected if the group mappings are wrong
+        """
+        # TODO check for event value especially for EV_ABS
+        try:
+            return event[0] in self.capabilities().keys() and event[1] in self.capabilities()[event[0]]
```
sezanzeb, line 41:
```suggestion
            return self.capabilities().keys().get(event[0], []).get(event[1]) is not None
```

sezanzeb, line 41:
```suggestion
            return event[1] in self.capabilities().get(event[0], [])
```

keymapper/injection/global_uinputs.py
sezanzeb: Have you checked the python-evdev code? This happens if `self.device is None`, so I guess `self._find_device()` returns None

jonasBoss: yes, i did check that code. It is calling `glob.glob(/dev/input/event*)` in order to get all  devices. However, this will only return the devices to which the user has read permissions. A simple fix would be to add the user to the input group. 

But we already know it is possible without adding the user to any group. If we can figure out why the python console creates a UInput with the correct permissions, that would be a better solution.

sezanzeb: Ok,

the systemd service is supposed to run as root anyway, so I guess it isn't really a problem

jonasBoss: No, it  doesn't seem to be a problem when running as a service. 
I updated the logic to handle this error in https://github.com/sezanzeb/key-mapper/pull/228/commits/f85ba92826af9c0e9852178031aac7fe0167e80c now we try to find the event* file once, and afterwards we ignore the error and assume the event can be handled by the device.

sezanzeb: If the service only ever runs with root permission once installed, is this really needed?

jonasBoss: We can write the events safely. The events get written to /dev/uinput not to /dev/input/event*.

we might not need to send the events anyway if /dev/input/event* is not available, but we should try to find the event* file at least once. 
UInput sleeps for 0.1s before it tries to find the event* file, but apparently that is not always enough. I did observe this on my system.

sezanzeb: > I did observe this on my system.

but did you run the code with `sudo`?

jonasBoss: okay, I just did some more testing. This is only a problem when running as a user who is in the input group. Apparently it takes more time to create the event* file if it is crated from a user.

When running as root this is not a problem, so I guess we are fine not catching Error

keymapper/injection/consumers/keycode_mapper.py
```diff
@@ -430,22 +433,41 @@ def handle_keycode(self, event, action, forward=True):
                 unreleased_entry = unreleased[type_and_code]
                 target_type, target_code = unreleased_entry.target_type_code
                 del unreleased[type_and_code]
-
+                
+                target_uinput = None
+                if isinstance(target_code, tuple):
+                    target_uinput = target_code[1]
+                    target_code = target_code[0]
```
sezanzeb, line 440:
```suggestion
                if isinstance(target_code, tuple):
                    target_code, target_uinput = target_code
```

When would target_code not be a tuple with the changes to `_map_keys_to_codes`?

sezanzeb, line 440: I think ideally it should be something like this, to keep the datastructure simple.

```py
target_uinput, target_type, target_code = unreleased_entry.target_type_code
```

and since this is a 3 tuple, `target_type_code` is not an appropriate name anymore. Maybe just `target`?

jonasBoss: Forwarded keycodes don't have a `target_uinput`, only a `target_code`

sezanzeb: Can we use `None` in `target_uinput` in that case? To have a consinstent data structure. Based on experience, handling different formats leads to bugs.

sezanzeb: Maybe it helps to establish a new class for this purpose

```py
class Target:
    def __init__(uinput, event_type, code, value):
        ...
```

You could use this in `_map_keys_to_codes`, and in `target_type_code` of `Unreleased`. That way there is a clear definition.

(since `type` is a builtin keyword, it need to be suffixed)

jonasBoss: > I think ideally it should be something like this, to keep the datastructure simple.

good idea, I just implemented it.
> The `keycode_mapper` was really hard to get at least somewhat readable.

Indeed, it is quite the challenge to see what is going on. I'm somewhat tempted to refactor the whole thing. Maybe breakout all the different logic paths into different functions and create a `event_handler_factory()` which will return the correct function. But not in this PR

sezanzeb: > I'm somewhat tempted to refactor the whole thing

understandable. It should probably be at least split into more functions. But yeah sure, lets do this in another PR.

jonasBoss: Okay, had a look how I might implement the Idea, but it does not seem worth the effort. I chose `tuple` on purpose because it is immutable if we replace it with a Target object we have to be careful not to mess up the mapping by accident. 
But I will keep the idea in mind if I end up refactoring `keycode_mapper`.

The changes I have done so far seem to work well. You can test them by changing a preset to something like this: 
```
{
    "mapping": {
        "1,3,1": ["BTN_A","gamepad"],
        "1,2,1": ["d","keyboard"],
        "1,4,1": ["c","keyboard"],
        "1,5,1+1,6,1": ["a","keyboard"],
        "1,7,1": ["b+BTN_2","keyboard"]
    }
}
```
And load it with `key-mapper-control` the GUI is probably broken.

sezanzeb: For immutability, I think making `def __setattr__()` throw an error would work.

I'll test it this weekend, thanks

sezanzeb:
Are you sure about the gui, wouldn't it be better to move this into the advanced editor? I'd really like to have the user interface as simple as possible for the simplest use cases.

jonasBoss:
No, I don't particularly like my adaptation for the GUI I just did it to see what it would take to program a GUI. I never used GTK before. 

However I don't think it is not so bad either there is just one thing I would change: I would replace the `target_input` with a dropdown menu. That way it is super simple to select the correct target. And in the future it is super simple to populate the dorpdown menu with user defined uinput devices. As a added bonus we can filter the selection list for the `symbol_input` to only contain keycodes the target actually supports.

sezanzeb:
I think I'll start working on the advanced editor over the holidays or something

jonasBoss:
A macro editor is a really good Idea. There are two more things I would like to add to the current UI: 

1. Filter the Gtk.EntryCompletion  for the symbol input
2. Validate the mapping before saving: Is there a way to get all keycodes a macro will trigger? I would like to check for every keycode to be in the target capabilities and give some visual feedback if the mapping is not valid. 

But take my changes as suggestions. Obviously they might not be compatible with the changes you make for the advanced editor. But I think those changes will be very helpful to users.

sezanzeb:
> Is there a way to get all keycodes a macro will trigger?

the macros track the used capabilities https://github.com/sezanzeb/key-mapper/blob/main/keymapper/injection/macros/macro.py#L251

this might break if variable support for key names is added: `set(a, KEY_A).key($a)`

keymapper/injection/global_uinputs.py
```diff
@@ -0,0 +1,108 @@
...
+                setattr(self, key, value)
+    
+    def capabilities(self):
+        return self.events
+        
+
+class GlobalUInputs:
+    """Manages all uinputs that are shared between all injection processes."""
+
+    def __init__(self, backend = True):
```
sezanzeb, line 69: `utils.is_service()` https://github.com/sezanzeb/key-mapper/blob/main/keymapper/utils.py#L240

keymapper/injection/global_uinputs.py
```diff
@@ -0,0 +1,139 @@
...
+        else:
+            self._uinput_factory = FrontendUInput
+
+    def prepare(self, force_service=False):
+        """Generate uinputs.
+
+        This has to be done in the main process before injections start.
+        """
+
+        # TODO: remove force_service we should find a way to patch keymapper.utils.is_service() in the tests
```
jonasBoss, line 99: @sezanzeb is there a good way to archive this? we need the `force_service` only in the tests because `is_service()` always returns `False`

jonasBoss: and there are some more problems where I would like to hear your thoughts: 

- the fake `UInput` class in `tests/test.py` does not work with `global_uinputs`: when running `test_injector.TestInjector.test_injector` it fails at line 641 with `5 != 11` even though the debug log shows all 11 events. I think all the forwarded events end up in the `uinput_write_history_pipe` but not the events mapped to a global UInput 
- `test_injector.TestInjector.test_store_permutations_for_macros` stops after a `ERROR consumer_control.py:118: The async_read_loop for "/dev/input/event10" stopped early` and never continues. I have no clue where I would start to debug this

sezanzeb: yes. Something like

```py
with patch.object(sys, "argv", ["key-mapper-service"]):
    ...
```

can probably be used in tests to make is_service return true

---

Is `patch_evdev` happening before `global_uinputs` creates its UInputs? 

---

The async_read_loop error might happen because `async_read_loop` of `InputDevice` in `test.py` finishes reading all events that are configured in the test. It has to stop in order for tests to continue. During runtime it would be an error, in tests it is expected.

sezanzeb: Put a few seconds of extra sleep before `history = read_write_history_pipe()` and see what happens. Maybe key-mapper is not handling the events fast enough and there is some bottleneck somewhere

sezanzeb: An alternative to `patch_evdev` with `evdev.UInput = UInput` might be to write those patches into the original `UInput` class:

```
evdev.UInput.write = uinput_write_patch
...
```

This might be more robust, but if that really works the way I think it works remains to be validated.

But if this is not the root cause then there is no priority here.

jonasBoss: > Is `patch_evdev` happening before `global_uinputs` creates its UInputs?

yes, I run `global_uinputs.prepare()` in the `cleanup()` function and the debug output contains a `grey_log(f"{(type, code, value)} written")` (line 435, test.py) line for each event

> I think you need to first use `push_events` with all the events you intend to read in your test, and then run the injector.

I did not touch the `test_store_permutations_for_macros` other than fixing a call to `mapping.change()` and the test does not read or write any events.

sezanzeb: I'll check it out in detail tomorrow evening

jonasBoss: > I'll check test_store_permutations_for_macros out in detail tomorrow evening

no need to, I found the error `injector._construct_capabilities()` is not used but the test relies on it to stop the injection before it starts.

sezanzeb: I wonder if there is anything we can do to make writing tests easier. Everything seems really complicated

keymapper/injection/macros/macro.py
```diff
@@ -287,7 +287,7 @@ async def run(self, handler):
                 if asyncio.iscoroutine(coroutine):
                     await coroutine
             except Exception as e:
-                logger.error(f'Macro "%s" failed: %s', self.code, e)
+                logger.error(f'Macro "{self.code}" failed: {e}')
```
sezanzeb, line 290: lazy logging is considered good style by pylint for performance reasons. Also see https://docs.python.org/3.8/howto/logging.html#optimization

sezanzeb: Well, except that logger.error will always do its thing and it doesn't apply here

keymapper/injection/global_uinputs.py
jonasBoss: yeah, it is quiet tedious to work on these tests.
> Put a few seconds of extra sleep before `history = read_write_history_pipe()` in `test_injector` and see what happens.

no difference at all.
I updated `write()` in the fake test UInput to print the pipe object it is writing to:

```
37622 22:46:41.522199 DEBUG consumer_control.py:83: Starting to listen for events from /dev/input/event20, fd 44
async_read_loop "Bar Device" "/dev/input/event20" InputEvent(1, 8, 1)
37622 22:46:41.522659 SPAM keycode_mapper.py:539: ((1, 8, 1)) ------------------- forwarding
(1, 8, 1) written to <multiprocessing.connection.Connection object at 0x7fb3f3d0cbb0>
async_read_loop "Bar Device" "/dev/input/event20" InputEvent(1, 9, 1)
37622 22:46:41.533072 SPAM keycode_mapper.py:516: ((1, 8, 1), (1, 9, 1)) -------- maps to macro (k(KEY_Q).k(w), keyboard)
37622 22:46:41.533208 DEBUG keycode_mapper.py:329: macro_sending (1, 101, 1)
(1, 101, 1) written to <multiprocessing.connection.Connection object at 0x7fb3f5b84fd0>
async_read_loop "Bar Device" "/dev/input/event20" InputEvent(1, 8, 0)
37622 22:46:41.543643 DEBUG keycode_mapper.py:329: macro_sending (1, 101, 0)
(1, 101, 0) written to <multiprocessing.connection.Connection object at 0x7fb3f5b84fd0>
37622 22:46:41.543816 SPAM keycode_mapper.py:463: ((1, 8, 0)) ------------------- forwarding release
(1, 8, 0) written to <multiprocessing.connection.Connection object at 0x7fb3f3d0cbb0>
37622 22:46:41.554053 DEBUG keycode_mapper.py:329: macro_sending (1, 102, 1)
(1, 102, 1) written to <multiprocessing.connection.Connection object at 0x7fb3f5b84fd0>
async_read_loop "Bar Device" "/dev/input/event20" InputEvent(1, 9, 0)
37622 22:46:41.554309 SPAM keycode_mapper.py:435: ((1, 9, 0)) ------------------- releasing macro
37622 22:46:41.554362 SPAM keycode_mapper.py:448: ((1, 9, 0)) ------------------- releasing key
37622 22:46:41.564650 DEBUG keycode_mapper.py:329: macro_sending (1, 102, 0)
(1, 102, 0) written to <multiprocessing.connection.Connection object at 0x7fb3f5b84fd0>
async_read_loop "Bar Device" "/dev/input/event20" InputEvent(3, 16, -1)
37622 22:46:41.564925 SPAM keycode_mapper.py:531: ((3, 16, -1)) ----------------- maps to (100, keyboard)
(1, 100, 1) written to <multiprocessing.connection.Connection object at 0x7fb3f5b84fd0>
async_read_loop "Bar Device" "/dev/input/event20" InputEvent(3, 16, 0)
37622 22:46:41.575298 SPAM keycode_mapper.py:454: ((3, 16, 0)) ------------------ releasing (100, keyboard)
(1, 100, 0) written to <multiprocessing.connection.Connection object at 0x7fb3f5b84fd0>
async_read_loop "Bar Device" "/dev/input/event20" InputEvent(1, 10, 1)
37622 22:46:41.585662 SPAM keycode_mapper.py:539: ((1, 10, 1)) ------------------ forwarding
(1, 10, 1) written to <multiprocessing.connection.Connection object at 0x7fb3f3d0cbb0>
async_read_loop "Bar Device" "/dev/input/event20" InputEvent(1, 10, 0)
37622 22:46:41.596140 SPAM keycode_mapper.py:463: ((1, 10, 0)) ------------------ forwarding release
(1, 10, 0) written to <multiprocessing.connection.Connection object at 0x7fb3f3d0cbb0>
async_read_loop "Bar Device" "/dev/input/event20" InputEvent(3124, 3564, 6542)
(3124, 3564, 6542) written to <multiprocessing.connection.Connection object at 0x7fb3f3d0cbb0>
37622 22:46:41.616714 ERROR consumer_control.py:118: The async_read_loop for "/dev/input/event20" stopped early
```

It looks like there are in fact two different pipes (at `0x7fb3f3d0cbb0` and `0x7fb3f5b84fd0`) I think this is currently my main issue, as it will cause a lot of tests to fail

sezanzeb: since `uinput_write_history` is a global variable it probably means there are two processes running. Hence the different identifiers.

It seems that the `b0` pipe is used for forwarding, and the `d0` pipe by the `keycode_mapper`

sezanzeb: You are using `global_uinputs.prepare` in `cleanup`, but `TestInjector` is using `quick_cleanup`. Could this be a contributing factor?

sezanzeb: If you have ideas for improving the tests in general, please throw them into https://github.com/sezanzeb/key-mapper/issues/246

jonasBoss: > You are using global_uinputs.prepare in cleanup, but TestInjector is using quick_cleanup. Could this be a contributing factor?

maybe but I think this is a bug in the overall design. Even in the main branch there are two pipes created but since all uinputs get created in the injector this is not a problem on main.

Using the traceback module we can see that the first pipe is created at the very beginning, this was previously never used. The second one gets created by the `import` statement in `test_injector.py`:
 ```
$ python tests/test.py test_injector.TestInjector.test_injector
creating pipe (<multiprocessing.connection.Connection object at 0x7f36ee8c5520>, <multiprocessing.connection.Connection object at 0x7f36ef3d9070>)
File "/home/jonas/dev/key-mapper/tests/test.py", line 111, in <module>
    for line in traceback.format_stack():
41672 00:12:12.313981 SPAM pipe.py:62: Creating new pipe for "/tmp/key-mapper-jonas/results"
41672 00:12:12.314752 SPAM pipe.py:62: Creating new pipe for "/tmp/key-mapper-jonas/commands"
cleanup
.
.
.
41672 00:12:12.401945 INFO groups.py:448: Found "Foo Device", "Foo Device 2", "Bar Device", "gamepad", "key-mapper Bar Device"
41672 00:12:12.402037 DEBUG global_uinputs.py:48: creating UInput device: 'key-mapper keyboard'
41672 00:12:12.402095 DEBUG global_uinputs.py:48: creating UInput device: 'key-mapper gamepad'
41672 00:12:12.402127 DEBUG global_uinputs.py:48: creating UInput device: 'key-mapper mouse'
creating pipe (<multiprocessing.connection.Connection object at 0x7f36ed5c0310>, <multiprocessing.connection.Connection object at 0x7f36ed5c03a0>)
File "/home/jonas/dev/key-mapper/tests/test.py", line 687, in <module>
    main()
File "/home/jonas/dev/key-mapper/tests/test.py", line 667, in main
    testsuite = unittest.defaultTestLoader.loadTestsFromNames(
File "/usr/lib/python3.9/unittest/loader.py", line 220, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
File "/usr/lib/python3.9/unittest/loader.py", line 220, in <listcomp>
    suites = [self.loadTestsFromName(name, module) for name in names]
File "/usr/lib/python3.9/unittest/loader.py", line 154, in loadTestsFromName
    module = __import__(module_name)
File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
File "<frozen importlib._bootstrap>", line 972, in _find_and_load_unlocked
File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
File "<frozen importlib._bootstrap>", line 972, in _find_and_load_unlocked
File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
File "<frozen importlib._bootstrap_external>", line 850, in exec_module
File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
File "/home/jonas/dev/key-mapper/tests/testcases/test_injector.py", line 70, in <module>
    from tests.test import (
File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
File "<frozen importlib._bootstrap_external>", line 850, in exec_module
File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
File "/home/jonas/dev/key-mapper/tests/test.py", line 111, in <module>
    for line in traceback.format_stack():
quick cleanup
41672 00:12:12.410838 DEBUG reader.py:196: Sending close msg to helper
```
I would guess this is caused by tests/test.py being run as `__main__` and then later imported as a module. The entry in https://docs.python.org/3/library/sys.html#sys.modules is only created once tests is loaded as a module, therefore we end up with two pipes

Comment (jonasBoss) [reply keymapper/injection/global_uinputs.py]: okay, turns out my guess was correct. This is an important lessons learned. I was able to fix it by moving the `if name == main` to the top and importing the file to itself:

```
import os
import sys

# make sure the "tests" module visible
sys.path.append(os.getcwd())
if __name__ == "__main__":
    # import this file to itself to make sure is not run twice and all global variables end up in sys.modules
    import tests.test
    tests.test.main()
    sys.exit(0)
```

Comment (jonasBoss) [reply keymapper/injection/global_uinputs.py]: and of course we are not alone with this problem:
https://stackoverflow.com/questions/13181559/importing-modules-main-vs-import-as-module

Comment (sezanzeb) [reply keymapper/injection/global_uinputs.py]: Nice! Please add the link to the stackoverflow post as a comment to the fix

We also have a circular import of some sort. The discoverer in test.py runs test files that import test.py again.

Comment (jonasBoss) [reply keymapper/injection/global_uinputs.py]: I think that is totally intended every test has to import `tests.test` in order to get access `quick_cleanup` and other stuff. But it is not circular until we import a `tests.testcases. ...` back into `tests.test` which we don't.

Comment (sezanzeb) [reply keymapper/injection/global_uinputs.py]: Ok. I was thinking, maybe the test utility functions like `quick_cleanup` should be moved into a separate file.

But if it doesn't break anything the way it is now I guess it's allright

**Comment by jonasBoss** [comment]
Apart from the migrations and the tests, this PR should support all features of the current key-mapper release. I think it is time you have a closer look at the code and see if you can find any issues. Also have a look at the GUI maybe you can think of a good way to integrate my changes with the advanced macro editor.

**Comment by sezanzeb** [comment]
cool, thank you! I think I'll finish the advanced editor this weekend and then look at it

**Comment by sezanzeb** [comment]
my new tests don't use the new `change` parameter yet, tests are failing. Should I just add `"keyboard"` there?

**Comment by jonasBoss** [comment]
Thank you for the info I was just about to look at the tests. I will wait with the GUI test until they work on main. 

I just added a target selector to the new GUI. It seems to work again.

**Comment by sezanzeb** [comment]
> I will wait with the GUI test until they work on main.

Tests on main are passing for me. What exactly is failing on your side?

File: inputremapper/migrations.py [hunk 9]
```diff
@@ -142,6 +142,37 @@ def _rename_config():
         shutil.move(old_config_path, CONFIG_PATH)
 
 
+def _add_target():
+    """add the target field to each preset mapping
+
+    Always use "keyboard" as target. This will break mappings which map codes not in the capabilities of keyboard.
+    We can not check code names or capabilities of the available targets due to circular imports
```
jonasBoss, line 149: Any Idea if we can do something about this? We currently can not access `system_mapping` so there is no good way to find a target with the right capabilities for the mapping.
also I am not sure if we can import `global_uinputs` and `parse`

jonasBoss:
Test are passing and I created new ones to cover the new code.
There is only the migration function which could be improved, I am not sure how.
And maybe we want to change the UI to better communicate what each target can send?

inputremapper/migrations.py
```diff
@@ -142,6 +142,37 @@ def _rename_config():
...
+    """
+    if not os.path.exists(get_preset_path()):
+        return  # don't execute if there are no presets
+
+    for preset in all_presets():
+        logger.info("Updating preset: %s", preset)
+        with open(preset, "r") as f:
+            preset_dict = json.load(f)
```
sezanzeb, line 157: we need to wrap usages of json.load to skip broken preset files

inputremapper/injection/global_uinputs.py
```diff
@@ -0,0 +1,143 @@
...
+            # "version": 1,
+            # "bustype": 3,
+            # "devnode": '/dev/uinput',
+            # "phys": 'py-evdev-uinput',
+        }
+        for key, value in defaults.items():
+            try:
+                setattr(self, key, kwargs[key])
+            except KeyError:
+                setattr(self, key, value)
```
sezanzeb, line 79: Isn't this equivalent to

```suggestion
    def __init__(self, *args, events=None, name="py-evdev-uinput", **kwargs):
        # see https://python-evdev.readthedocs.io/en/latest/apidoc.html#module-evdev.uinput
        self.events = events
        self.name = name
```

?

data/input-remapper.glade
```diff
@@ -873,12 +873,12 @@ Gives your keys back their original function</property>
                               </packing>
                             </child>
                             <child>
-                              <object class="GtkLabel">
+                              <object class="GtkComboBox" id="target-selector">
                                 <property name="visible">True</property>
                                 <property name="can-focus">False</property>
                               </object>
```
sezanzeb, line 879: Added tooltip, homogeneous positioning of buttons and dropdown: https://gist.github.com/sezanzeb/e542fc25deb2e1bb894959827abc1d22

sezanzeb:
in editor.py there is `custom_mapping.change(new_key=key, symbol=symbol, previous_key=previous_key)` without a target parameter

sezanzeb (changes requested):
I like it, thanks a lot

inputremapper/migrations.py
```diff
@@ -155,6 +186,9 @@ def migrate():
     if v < pkg_resources.parse_version("1.3.0"):
         _rename_config()
 
+    if v < pkg_resources.parse_version("1.3.1"):
```
sezanzeb, line 189: This would be in 1.4.0

```suggestion
    if v < pkg_resources.parse_version("1.4.0"):
```

inputremapper/migrations.py
```diff
@@ -142,6 +142,37 @@ def _rename_config():
...
+    """add the target field to each preset mapping
+
+    Always use "keyboard" as target. This will break mappings which map codes not in the capabilities of keyboard.
+    We can not check code names or capabilities of the available targets due to circular imports
+    """
+    if not os.path.exists(get_preset_path()):
+        return  # don't execute if there are no presets
+
+    for preset in all_presets():
+        logger.info("Updating preset: %s", preset)
```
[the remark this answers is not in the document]
jonasBoss: That would require extra logic to detect if something changed. I do print exactly which key was changed inside the inner for loop. I could include the preset information there and remove this line.

inputremapper/gui/editor/autocompletion.py
```diff
@@ -371,6 +376,10 @@ def update(self, *_):
             self.list_box.insert(label, -1)
             label.show_all()
 
+    def _update_codes(self):
+        target = self.target_selector.get_active_id()
+        self._codes = global_uinputs.get_uinput(target).capabilities()[EV_KEY]
```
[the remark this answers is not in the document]
jonasBoss: Good Idea, and it works as expected.

inputremapper/injection/global_uinputs.py
jonasBoss: yeah it is. Not sure why I made it this way.

inputremapper/migrations.py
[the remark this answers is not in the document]
sezanzeb: yes lets do that instead

sezanzeb:
If I correctly remember, in https://github.com/sezanzeb/input-remapper/issues/221 the issue with wayland was that shift + a would have to inject the shift into the same uinput as the mapping is injected to.

image

this is supposed to write lowercase `b`s in wayland, because the shift key should be released via `key(shift_L)`.

sezanzeb:
nevermind, this makes it work:

image

Is this reade to be merged now?

inputremapper/migrations.py
sezanzeb: why can't we import global_uinputs? Where is the circular import?

sezanzeb: seriously, if you just check if there is a BTN_ written somewhere in the mapping and use gamepad instead for that it's going to be fine probably

jonasBoss:
Wait, why does it work? it should not:

I am not sure if I understand this correctly, but I did some testing and it looks like a bug in X11. Wayland behaves exactly as I would it expect to.
### Wayland
image
As expected we get a capital B as the `Shift_L` from the original keyboard is held down. The `Shift_L` form the macro has no effect since it is already released when the `b` event is sent. (The macro is a sequence, not a combination)
When not using a macro we get the same behavior:
image
Wayland makes perfectly sense to me. what does not make sense it X11

## X11
I did the exact same tests:
image
Why does X11 care about the `Shift_L` which was sent before the `b` event? Is this by design or a bug? (possibly a race condition)
When not sending the `Shift_L` in the macro everything works again:
image

sezanzeb:
The problem is that in wayland it was not possible to map `shift_l + a` to a lowercase `b`.

By mapping shfit_l to (shift_l, keyboard) the macro can release shift_l.

There should be a macro called `key_up` for this purpose, instead of using `k(shift_l)` to release, but that is a different story. There is also a feature request for this macro function already: https://github.com/sezanzeb/input-remapper/issues/198#issuecomment-958788175

jonasBoss:
Okay, your solution seems to avoid the Bug in X11. But it is not super intuitive maybe we can handle combinations differently in general:
 1. we detect that a combination triggered a mapping
 2. send release events to the forwarded device (for all keys involved in the combination)
 3. do whatever the mapping says
 
we need a macro function which allows us to wrap a sub macro by inside a key event maybe like this:

```
wrap(key(a).key(b), Shift_L)

events generated:
Shift_L down
  a down
  a up
  b down
  b up
Shift_L up

output:
AB
```

inputremapper/migrations.py
jonasBoss: No, i think that would work. The problem is that we can not import system_mapping, DEFAULT_UINPUTS and global_uinputs only contain the key-codes, not all the different key-names which are in the preset files?

sezanzeb:
this would be `modify(Shift_L, key(a).key(b))`

sezanzeb:
> - we detect that a combination triggered a mapping
> - send release events to the forwarded device (for all keys involved in the combination)
> - do whatever the mapping says

could this be a breaking change for some? could there be side effects? Maybe some people dont want their key to be released for whatever reason. lets discuss this separately

inputremapper/migrations.py
sezanzeb: just try/catch it for json.decoder.JSONDecodeError

```

        with open(preset, "r") as f:
            preset_dict = json.load(f)

File: inputremapper/migrations.py (hunk 9 above)
Comment (sezanzeb) [reply inputremapper/migrations.py:149]: ah right.

how about something like

```
>>> import evdev
>>> evdev.ecodes.ecodes["KEY_A"]
30

sezanzeb: If `migrate()` is called in `bin/input-remapper-gtk` and `bin/input-remapper-control` instead of `config` I think there won't be a circular import anymore

jonasBoss: how does the autoloading process work exactly? Can we be sure migrate() is called before any presets autoload?

sezanzeb: autoloading uses input-remapper-control: https://github.com/sezanzeb/input-remapper/blob/main/data/99-input-remapper.rules

Just make sure to import it early and call it before importing the other modules.

jonasBoss: seems to work now

sezanzeb:
I'll merge it if you say its ready

jonasBoss:
Have a look at my last commit, If that is okay I would say the PR is ready.

sezanzeb:
This really makes sense, thanks for your work

sezanzeb:
Should there be also a "mousepad" and "tablet" uinput by default?

tablet: https://github.com/torvalds/linux/blob/master/include/uapi/linux/input-event-codes.h#L400

mousepad: https://github.com/torvalds/linux/blob/master/include/uapi/linux/input-event-codes.h#L854

jonasBoss:
it is super simple to add those. I think it would not hurt.

PR #230, opened 2021-12-10
Independent uinput proposals

sezanzeb:


sezanzeb:
@jonasBoss, check out the GlobalUInputs class, what do you think of that?

keymapper/injection/global_uinputs.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+import evdev
+from evdev.ecodes import EV_KEY
+
+from keymapper.utils import DEV_NAME, is_keyboard_code
+
+
+class GlobalUInputs:
+    """Manages all uinputs that are shared between all injection processes."""
```
jonasBoss, line 29: Seems quite sensible, and as you already said, it avoids passing around the uinput_devices dict.

keymapper/injection/global_uinputs.py
```diff
@@ -0,0 +1,68 @@
...
+    def __init__(self):
+        self.keyboard = None
+        self.gamepad = None
```
jonasBoss, line 32: I think we should internally manage the devices in a dict and not hardcode them. That way we can create new devices or remove unwanted devices during runtime.

sezanzeb: yes

keymapper/injection/global_uinputs.py
sezanzeb: great. Lets continue on your branch anyway, maybe you can copy it over or something

keymapper/injection/consumers/consumer.py
```diff
@@ -54,12 +59,21 @@ def is_enabled(self):
 
     def write(self, key):
         """Shorthand to write stuff."""
-        self.context.uinput.write(*key)
-        self.context.uinput.syn()
+        uinput = global_uinputs.get_appropriate_uinput(key)
```
jonasBoss, line 62: a consumer should be able to send different events to different global UInputs.

keymapper/injection/global_uinputs.py
```diff
@@ -0,0 +1,68 @@
...
+        self.keyboard = evdev.UInput(
+            name="key-mapper keyboard",
+            phys=DEV_NAME,
+            events={evdev.ecodes.EV_KEY: keys},
+        )
+
+    def get_appropriate_uinput(self, key):
+        """Where to inject the event to.
+
+        Or None if there is no appropriate device.
```
jonasBoss, line 55: This seems problematic:
1. There might be different UInput devices which can handle the same events. 
2. We need to know the mapping of the group from which the event is coming. This mapping has to include the Target device. If that is the case this method becomes way simpler: ` def get_appropriate_uinput(self, target): return devices[target]`

keymapper/injection/consumers/consumer.py
```diff
@@ -54,12 +59,21 @@ def is_enabled(self):
...
+
+        uinput.write(*key)
+        uinput.syn()
 
     def forward(self, key):
         """Shorthand to forward an event."""
-        self.forward_to.write(*key)
+        uinput = global_uinputs.get_appropriate_uinput(key)
+        if uinput is None:
+            uinput = self.forward_to
```
jonasBoss, line 73: I am not sure if we actually want to catch every possible event with a global UInput as long as a single event of that group is mapped. This seems quite confusing. I would expect any event which is not mapped to end up in the forwarded device.

keymapper/injection/global_uinputs.py
jonasBoss: yea, I will see what I can copy/merge

keymapper/injection/global_uinputs.py
sezanzeb: > There might be different UInput devices which can handle the same events.

you are right. Both gamepads and graphics tablets use ABS_X and ABS_Y

> This mapping has to include the Target device

any ideas on how it has to be specified? Is there a default fallback if it is not specified?

keymapper/injection/consumers/consumer.py
sezanzeb: I see

For a mapping of `shift_l + a: b` we definitely have to write all shift_l and a events to the global_uinputs, so that `e(EV_KEY, KEY_LEFTSHIFT, 0)` can be used to release the shift.

keymapper/injection/global_uinputs.py
jonasBoss: I think we should add the target information directly to the presets in json it would look something like this:
```
{
    "mapping": {
        "1,2,1": ["BTN_1", "gamepad"],
        "1,3,1": ["BTN_2", "gamepad"], 
        "1,4,1": ["KEY_A", "keyboard"]
    }
}
```
The fallback would be not mapping the event and forwarding it. The only events which have no target, are those which are not mapped.

sezanzeb: > Maybe we can create some script which converts existing presets.

For example there is also `migrate_path` and backwards compatibility in `split_key`. We can add a migration, yes

How is this configured in the gui?

sezanzeb: And how is it configured for makros? `key(KEY_A).key(BTN_1)`

sezanzeb: maybe `use("gamepad").key(BTN_1).key(BTN_2).key(BTN_3)`?

If an old preset is migrated, which uinput names are used. 

Is this an advanced feature that requires editing the json config?

PR #232, opened 2021-12-13
Refactored Migrations

jonasBoss:
I moved the migrations to a common place and implemented a simple system to keep track of the applied migrations.
I replaced the backwards compatibility in split_key() with a migration and updated the migration related tests.

This should clean up some classes and provide a better starting point for future migrations.

### DIFF
--- a/keymapper/migrations.py
+++ b/keymapper/migrations.py
@@ -69,17 +69,20 @@ def _preset_path():
     Move existing presets into the new subfolder "presets"
     """
     new_preset_folder = os.path.join(CONFIG_PATH, "presets")
-    if not os.path.exists(get_preset_path()) and os.path.exists(CONFIG_PATH):
-        logger.info("Migrating presets from < 0.4.0...")
-        groups = os.listdir(CONFIG_PATH)
-        mkdir(get_preset_path())
-        for group in groups:
-            path = os.path.join(CONFIG_PATH, group)
-            if os.path.isdir(path):
-                target = path.replace(CONFIG_PATH, new_preset_folder)
-                logger.info('Moving "%s" to "%s"', path, target)
-                os.rename(path, target)
-        logger.info("done")
+    if os.path.exists(get_preset_path()) or not os.path.exists(CONFIG_PATH):
+        return
+    
+    logger.info("Migrating presets from < 0.4.0...")
+    groups = os.listdir(CONFIG_PATH)
+    mkdir(get_preset_path())
+    for group in groups:
+        path = os.path.join(CONFIG_PATH, group)
+        if os.path.isdir(path):
+            target = path.replace(CONFIG_PATH, new_preset_folder)
+            logger.info('Moving "%s" to "%s"', path, target)
+            os.rename(path, target)
+            
+    logger.info("done")
 
 
 def _mapping_keys():
@@ -93,7 +96,8 @@ def _mapping_keys():
             mapping = copy.deepcopy(preset_dict["mapping"])
             for key in mapping.keys():
                 if key.count(",") == 1:
-                    preset_dict["mapping"][key + ",1"] = preset_dict["mapping"].pop(key)
+                    preset_dict["mapping"][f"{key},1"] = preset_dict["mapping"].pop(key)
+                    
         with open(preset, "w") as file:
             json.dump(preset_dict, file, indent=4)
             file.write("\n")
--- a/keymapper/migrations.py
+++ b/keymapper/migrations.py
@@ -31,6 +31,7 @@ from pathlib import Path
 from keymapper.logger import logger
 from keymapper.paths import get_preset_path, mkdir, CONFIG_PATH
 
+
 def all_presets():
     p = Path(get_preset_path())
     presets = []
@@ -78,7 +79,10 @@ def _preset_path():
                 os.rename(path, target)
         logger.info("done")
 
+
 def _mapping_keys():
+    if not os.path.exists(get_preset_path()):
+        return # don't execute if there are no presets
     for preset in all_presets():
         preset_dict = {}
         with open(preset, "r") as file:
@@ -94,6 +98,8 @@ def _mapping_keys():
 
 def _update_version(version):
     config_path = os.path.join(CONFIG_PATH, "config.json")
+    if not os.path.exists(config_path):
+        return
     with open(config_path, "r") as file:
         config = json.load(file)
     config["migration"] = version
--- a/keymapper/migrations.py
+++ b/keymapper/migrations.py
@@ -34,6 +34,7 @@ from keymapper.paths import get_preset_path, mkdir, CONFIG_PATH
 
 
 def all_presets():
+    """All presets for all groups as list"""
     preset_path = Path(get_preset_path())
     presets = []
     for folder in preset_path.iterdir():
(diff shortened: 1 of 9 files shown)


keymapper/migrations.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def all_presets():
+    p = Path(get_preset_path())
+    presets = []
+    for folder in p.iterdir():
+        if folder.is_dir():
+            for preset in folder.iterdir():
+                if preset.suffix == ".json":
+                    presets.append(preset)
```
sezanzeb, line 42: https://medium.com/swlh/return-early-pattern-3d18a41bba8

https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html

```suggestion
    for folder in p.iterdir():
        if not folder.is_dir():
            continue
                
        for preset in folder.iterdir():
            if preset.suffix == ".json":
                presets.append(preset)
```

keymapper/migrations.py
```diff
@@ -0,0 +1,118 @@
...
+        logger.info("Migrating presets from < 0.4.0...")
+        groups = os.listdir(CONFIG_PATH)
+        mkdir(get_preset_path())
+        for group in groups:
+            path = os.path.join(CONFIG_PATH, group)
+            if os.path.isdir(path):
+                target = path.replace(CONFIG_PATH, new_preset_folder)
+                logger.info('Moving "%s" to "%s"', path, target)
+                os.rename(path, target)
+        logger.info("done")
```
sezanzeb, line 80: Also here, "return early"/"guard" pattern helps to make things cleaner in the long term. Like you already did in `_mapping_keys` correctly

```suggestion
    if os.path.exists(get_preset_path()) or not os.path.exists(CONFIG_PATH):
        return
    
    logger.info("Migrating presets from < 0.4.0...")
    groups = os.listdir(CONFIG_PATH)
    mkdir(get_preset_path())
    for group in groups:
        path = os.path.join(CONFIG_PATH, group)
        if os.path.isdir(path):
            target = path.replace(CONFIG_PATH, new_preset_folder)
            logger.info('Moving "%s" to "%s"', path, target)
            os.rename(path, target)
            
    logger.info("done")
```

I was told that it is good style to put newlines after if/for/while statements

keymapper/migrations.py
```diff
@@ -0,0 +1,118 @@
...
+def migrate():
+    v = version()
+    if v == 0:
+        _config_suffix()
+        v = 1
+    if v == 1:
+        _preset_path()
+        _mapping_keys()
+        v = 2
+        _update_version(2)
```
sezanzeb, line 118: would it make sense to use a semver string here?

`version.parse(v) < version.parse("1.0.0")`

https://stackoverflow.com/a/11887885/4417769

We would still store the version in the config to remember up to which point things were migrated. We just use a different format for it to see with which version this migration corresponds, to improve readability. So that we, as developers, know when this migration was supposed to run.

keymapper/migrations.py
```diff
@@ -0,0 +1,118 @@
...
+    for preset in all_presets():
+        preset_dict = {}
+        with open(preset, "r") as file:
+            preset_dict = json.load(file)
+        if "mapping" in preset_dict.keys():
+            mapping = copy.deepcopy(preset_dict["mapping"])
+            for key in mapping.keys():
+                if key.count(",") == 1:
+                    preset_dict["mapping"][key + ",1"] = preset_dict["mapping"].pop(key)
+        with open(preset, "w") as file:
```
sezanzeb, line 95: formatstrings are considered good style (and apparently adding newlines after if-statements and loops as well, according to a colleague who has been into clean code for a long time)

```suggestion
                    preset_dict["mapping"][f"{key},1"] = preset_dict["mapping"].pop(key)
                    
        with open(preset, "w") as file:
```

tests/testcases/test_mapping.py
```diff
@@ -40,7 +40,7 @@ def tearDown(self):
 
     def test_split_key(self):
         self.assertEqual(split_key("1,2,3"), (1, 2, 3))
-        self.assertEqual(split_key("1,2"), (1, 2, 1))
+        # self.assertEqual(split_key("1,2"), (1, 2, 1))
```
sezanzeb, line 43: lets remove this line

sezanzeb (changes requested):
Thanks a lot! This makes a lot of sense.

I just added mostly style comments. Also please make sure to use `black` on the code once (This should probably be documented in development.md)

keymapper/migrations.py
```diff
@@ -0,0 +1,118 @@
...
+import copy
+
+from pathlib import Path
+
+from keymapper.logger import logger
+from keymapper.paths import get_preset_path, mkdir, CONFIG_PATH
+
+
+def all_presets():
+    p = Path(get_preset_path())
```
sezanzeb, line 36: using written-out variable names is better style

probably

```suggestion
    preset_path = Path(get_preset_path())
```

sezanzeb:
Were tests passing on your computer?

I caused some race conditions in the tests that make some of them fail on my laptop

keymapper/migrations.py
sezanzeb: Also, if someone makes a clean install, what will v be initialized with?

If we use a semver string here, we can place the current version at installation time into `INITIAL_CONFIG`

jonasBoss: makes totally sense. For a clean install we should not need any migrations. So probably initialize with the current version.

jonasBoss:
yes, the test are passing on my pc.

sezanzeb (changes requested):
Test coverage looks good, and the new tests also make sense. Migrations work on my computer correctly. Well done :)

I'd like to release 1.2.2 first (probably tomorrow) and then merge this, if that is ok, because the commits since the last release seem to be stable enough now.

I usually add short docstrings to all of my functions. They don't have to be fancy or anything, just a one-liner sentence on what this function will do. I'd appreciate if you could add a few to those that are missing one.

keymapper/migrations.py
```diff
@@ -0,0 +1,133 @@
...
+        json.dump(config, file, indent=4)
+
+
+def migrate():
+    v = config_version()
+    if v < version.parse("1.2.1"):
+        _config_suffix()
+        _preset_path()
+        _mapping_keys()
+        _update_version()
```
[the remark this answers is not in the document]
jonasBoss: the other way around: First migrate to 0.4.0 

Since this is probably going to be implemented in 1.2.3 we should check `if v < 1.2.3` but this would make it difficult to test as long as we are on 1.2.2 so I suggest to introduce a 1.2.3.dev1 version with this PR

keymapper/migrations.py
```diff
@@ -0,0 +1,133 @@
...
+    if not os.path.exists(config_path):
+        return version.parse("0.4.0")
+
+    with open(config_path, "r") as file:
+        config = json.load(file)
+
+    if "version" in config.keys():
+        return version.parse(config["version"])
+
+    return version.parse("0.4.0")
```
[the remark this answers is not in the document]
jonasBoss: makes sense

jonasBoss:
I added the docstrings and updated migrations() it now checks `if config_version < 1.2.3.dev1`

sezanzeb:
> I added the docstrings and updated migrations() it now checks if config_version < 1.2.3.dev1

I think it needs to be `1.2.3-dev1` in that case (see https://semver.org)

I'm making the 1.2.2 release now, then merge this, and then make a `1.2.3-dev1` release

sezanzeb:
I'll release it as part of 1.2.2, I don't have a reason to believe why it would not be stable

Issue #253, opened 2022-01-11
failing tests in main branche


jonasBoss:
> I will wait with the GUI test until they work on main.

Tests on main are passing for me. What exactly is failing on your side?

_Originally posted by @sezanzeb in https://github.com/sezanzeb/input-remapper/issues/228#issuecomment-1009971301_

```
======================================================================
FAIL: test_autocomplete_function (testcases.test_gui.TestAutocompletion)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jonas/dev/key-mapper/tests/testcases/test_gui.py", line 1852, in test_autocomplete_function
    self.add_mapping_via_ui(Key(1, 99, 1), "")
  File "/home/jonas/dev/key-mapper/tests/testcases/test_gui.py", line 391, in add_mapping_via_ui
    self.assertFalse(self.toggle.get_active())
AssertionError: True is not false

======================================================================
FAIL: test_autocomplete_key (testcases.test_gui.TestAutocompletion)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jonas/dev/key-mapper/tests/testcases/test_gui.py", line 1809, in test_autocomplete_key
    self.add_mapping_via_ui(Key(1, 99, 1), "")
  File "/home/jonas/dev/key-mapper/tests/testcases/test_gui.py", line 391, in add_mapping_via_ui
    self.assertFalse(self.toggle.get_active())
AssertionError: True is not false

======================================================================
FAIL: test_close_autocompletion (testcases.test_gui.TestAutocompletion)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jonas/dev/key-mapper/tests/testcases/test_gui.py", line 1873, in test_close_autocompletion
    self.add_mapping_via_ui(Key(1, 99, 1), "")
  File "/home/jonas/dev/key-mapper/tests/testcases/test_gui.py", line 391, in add_mapping_via_ui
    self.assertFalse(self.toggle.get_active())
AssertionError: True is not false

======================================================================
FAIL: test_cycling (testcases.test_gui.TestAutocompletion)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jonas/dev/key-mapper/tests/testcases/test_gui.py", line 1923, in test_cycling
    self.add_mapping_via_ui(Key(1, 99, 1), "")
  File "/home/jonas/dev/key-mapper/tests/testcases/test_gui.py", line 391, in add_mapping_via_ui
[...]

PR #257, opened 2022-01-14
the ui no longer records keycodes if the preset is active

jonasBoss:
since there can be multiple evdev devices in the same group it was possible to record keycodes even if the injection was active. This was confusing behaviour.

### DIFF
--- a/inputremapper/gui/editor/editor.py
+++ b/inputremapper/gui/editor/editor.py
@@ -32,7 +32,7 @@ from inputremapper.gui.custom_mapping import custom_mapping
 from inputremapper.key import Key
 from inputremapper.logger import logger
 from inputremapper.gui.reader import reader
-from inputremapper.gui.utils import CTX_KEYCODE, CTX_WARNING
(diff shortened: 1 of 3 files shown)


sezanzeb:
gui tests are no longer passing

jonasBoss:
On my system they are passing

sezanzeb:
ah, I forgot about https://github.com/sezanzeb/input-remapper/issues/253

jonasBoss:
maybe you should add a hint:
https://github.com/sezanzeb/input-remapper/blob/main/readme/development.md#tests

sezanzeb:
It is now possible to run tests via PyCharms built in tools. There is a "All Tests" configuration now.

sezanzeb:
Can you please adjust `test_can_modify_mapping` to check if it won't allow focusing the toggle

jonasBoss:
that test was actually broken, it returned early and never checked for the "Stop injection" text. It should have failed for this PR

sezanzeb:
thanks!

Issue #229, opened 2021-12-10
Combinations involving Modifiers on Wayland. Modifier key not released.


petrstepanov:
Dear @sezanzeb, thank you for this great remapping tool. I love the UI and scripting approach. I believe I found an issue with cancelling the modifier key for combination of keys.

The "Combinations" section of the documentation says that the modifier key in combination can be released by writing the modifier one more time. Here is the picture I am referring to:

<img src="https://github.com/sezanzeb/key-mapper/raw/main/readme/combination_workaround.png" />

Scenario: user presses `Shift_L + p`. 
Expected result: lowercase `a` is written.
 
I did test this scenario on Xorg and Wayland and it only seems to work on Xorg. On Wayland the uppercase `A` is written instead. Please refer to the screenshot below:

<img src="https://i.imgur.com/lQ88jAV.png"/>

Here is my environment:
```
> uname -a
Linux fedora 5.15.6-200.fc35.x86_64 #1 SMP Wed Dec 1 13:41:10 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

> key-mapper-control --version
key-mapper 1.2.1 b23e6d5340bb3c0e6337196d13af5c69d6b6540d https://github.com/sezanzeb/key-mapper
python-evdev 1.4.0

> echo $XDG_SESSION_TYPE
wayland
```

I skimmed through the "Issues" and "Discussions" pages and did not find the issue documented. I also tried installing key-mapper from `main` and `key-mapper-devices` branches.

Thanks again for your time and efforts. Looking forward to hearing from you!

sezanzeb:
Hi,

to verify if the installation from `key-mapper-devices` worked correctly, you can print the commit hash that is currently being used with this:

```
key-mapper-control --version
key-mapper 1.2.1 cb39d39eeb74a89b0337bf8e01e857cddfed91db https://github.com/sezanzeb/key-mapper
```

petrstepanov:
Thank you. My bad. I must have missed something. Just re-tested and ensured the behavior is correct. Any estimate when this will be merged to the main branch?

sezanzeb:
There is some work being done here: https://github.com/sezanzeb/key-mapper/pull/228

@JonasBoss we might want to first get the global uinputs to work and then work on arbitrary configurable uinputs in a separate PR to get this released faster, what do you think?

jonasBoss:
Yes, that is pretty much my current idea. First get all the current features working with a few predefined global uinputs.
[...]

Issue #183, opened 2021-09-17
Space Cadet Shift


nan0tube:
Any idea if Space Cadet Shift <sup>[1](#fn1)</sup><sup>[2](#fn2)</sup> can be mapped/implemented?

<a name="fn1">1</a>: [Space Cadet Shift - QMK](https://beta.docs.qmk.fm/using-qmk/advanced-keycodes/feature_space_cadet)
<a name="fn2">2</a>: [A Modern Space Cadet / Steve Losh](https://stevelosh.com/blog/2012/10/a-modern-space-cadet/#s17-shift-parentheses)

sezanzeb:
I think I just came up with a way to implement this. But I need some time to think about the best way to make this configurable. 

How would you name a macro function that
- does exactly what is described above, but configurable to write other keys than `(` and `)`
- works like `if the key is still the most recently pressed key then ... else ...`
- works like `if the key has been combined with any other key then ... else ...`
- works like `if this key has been combined with this keycode + value then ... else ...`

nan0tube:
I just found out about key-mapper, so may not be able to come up with a name that is consistent with other parts of the project. Nevertheless here's my shot at it:
```
mod(mod_key, combined_key, <mod_key+combined_key event>, <mod_key+!combined_key event>, <mod_key only event>)
```

sezanzeb:
what about

```
ifTab(a, b, 100)
# and even possibly at some point
ifDoubleTab(a, b, 100)
```
- if tabbed write a
- otherwise (if combined with another key or if too much time passes) b
- the number would be the timeframe in which the key has to be released in ms, 0 (default) to make it infinitely large

nan0tube:
Perhaps you meant `if_tap`? In any case this looks way more elegant!

sezanzeb:
yes, thanks. Ok

sezanzeb:
I want to rewrite the architecture a little bit for this, it will take a little while. I'll let you know when there is something to try out

sezanzeb:
Please try the https://github.com/sezanzeb/key-mapper/tree/if_tap branch

writing opening and closing brackets is a bit complicated if they are only accessible via modifiers. On my keyboard it would be:

`m(KEY_LEFTSHIFT, k(8))`

So for example this macro should write a "(" if the key is released within 1 second. If the key is combined with any other key or if you wait for too long it should write a shift

[...]

Issue #7, opened 2020-12-26
Not running when starting via GUI 


clausbertels:
Hi, I'd like to quickly say that this is great! I reckon many people have been waiting for this and it will advance many distros in becoming more user friendly. Great work!

I'm on Elementary OS 5.1. I've been fiddling with xbindkeys and logiops to get my MX Keys and MX Master 3 to play nicely with my system. It's a lot of work to get things to function well. Today I discovered your app and hoped it could bring some ease to my headaches.

## The Issue

I couldn't install the 0.4.0.deb with Eddy, so I tried directly from source. This seemed to install fine, but when I launch via applications, it pops up in the Dock for 5 seconds and then quits. No GUI. I ran `sudo key-mapper-gtk -d`

```
Loaded config from "/home/claus/.config/key-mapper/config.json"
Writing "/home/claus/.config/key-mapper/xmodmap.json"
INFO: logger.py, line 114, key-mapper 0.4.0
WARNING: logger.py, line 121, Debug level will log all your keystrokes! Do not post this output in the internet if you typed in sensitive or private information with your device!
DEBUG: logger.py, line 126, pid 11336
DEBUG: data.py, line 86, Found data at "/usr/share/key-mapper"

(key-mapper-gtk:11336): IBUS-WARNING **: 15:47:40.554: The owner of /home/claus/.config/ibus/bus is not root!

(key-mapper-gtk:11336): IBUS-WARNING **: 15:47:40.584: Unable to connect to ibus: Unexpected lack of content trying to read a line
DEBUG: getdevices.py, line 93, Discovering device paths
SPAM: getdevices.py, line 116, Found "LogiOps Virtual Input", /dev/input/event4, 
SPAM: getdevices.py, line 116, Found "Logitech MX Keys Mac", /dev/input/event3, usb-0000:02:00.0-4
SPAM: getdevices.py, line 116, Found "Logitech MX Master 3", /dev/input/event2, usb-0000:02:00.0-4
INFO: getdevices.py, line 179, Found "LogiOps Virtual Input", "Logitech MX Keys Mac"
DEBUG: presets.py, line 144, The newest preset is "Logitech MX Keys Mac", "new preset"
DEBUG: window.py, line 422, Selecting device "Logitech MX Keys Mac"
DEBUG: window.py, line 241, "Logitech MX Keys Mac" presets: "new preset"
DEBUG: window.py, line 469, Selecting preset "new preset"
[...]

PR #263, opened 2022-01-17
Refactored injection

jonasBoss:
### New Features
 - Multiple Trigger points for mapping `EV_ABS` or `EV_REL` to buttons
 - Simultaneous mapping of `EV_ABS` or `EV_REL` to buttons and other analog axis
 - All `EV_ABS` event codes can be mapped to axis and or buttons
   - No limitation for mapping only joystick to mouse or scroll wheel
 - Individually (per mapping) configurable mapping parameters
   - trigger point, deadzone, gain, expo

### Changes
 1. replaced all `Consumer` classes with multiple `MappingHandler` classes
 2. injecting release events to the forwarded UInput before combinations trigger
 3. reworked macro if_singel to work with new injection architecture
 4. renamed `Mapping` class to `Preset`
 5. created new `Mapping` class responsible for mapping a event or event combination to an output action
 6. moved multiple configuration classes to new `configs` directory
 7. renamed `ConsumerControl` to `EventReader`
 8. replaced all `evdev.InputEvent` objects with custom `inputremapper.InputEvent` objects
 9. Replaced the `Key` class with a `EventCombination` class of type `Tuple[InputEvent]`

### TODO

- [x] Integrate the new `Mapping` class into the `Preset`
- [x] Implement Setting for [automatic key release](https://github.com/sezanzeb/input-remapper/discussions/305#discussioncomment-2204575) see also #229
- [ ] make the `Editor`-Part of the GUI interact with the new `Mapping` instead of the `Preset`
- [ ] Patch the GUI to create `Mapping` objects for Joystick to Mouse
- [x] modify `Injector._grab_device` to purely rely on mappings, no longer use `maps_joystick()`
- [x] expose the `Mapping` in the preset files
- [x] rework the `mapping_parser`
- [x] migrations
- [x] make sure to release any keys, center any axis before stopping the injection
- [ ] Tests, Tests, Tests, Tests
- [ ] Update documentation

### Future (probably not this PR)
 - Update GUI To expose the new functionality
 - Implement `RelToAbsHandler`
 - Implement `RelToRelHandler`
 - Implement `AbsToAbsHandler`
 - Implement Offset for axis to axis handlers


inputremapper/injection/context.py
```diff
@@ -20,13 +20,54 @@
...
+    in this example b would be the common key
+
+    Parameters
+    ----------
+    keys : List[Key]
+        the list which needs ordering
+    key: Key
+        the Key all members of Keys have in common
+    """
+    # TODO
```
jonasBoss, line 56: This would be the code responsible for making combinations behave correctly.
I am open for suggestions, especially what to do about `a+b+c` and `b+d+e` we should at least make sure it behaves always the same.

sezanzeb: what

jonasBoss: If the Key `b` ((1,48,1),) is pressed the event gets sent to all `MappingHandlers` which need to know about this key. 

> A mapping handler handles a single mapping in the preset e.g. `"1,30,1": ["a" , "keyboard"]` or `"1,30,1+1,31,1": ["BTN_LEFT" , "mouse"]`. It can be hierarchic (`KeysToKey`) or not (`HirarchyHandler`). For each event only one hierarchic handler can be notified, but all non hierarchic handlers will be notified.
> To achieve that we order all hierarchic handlers with `order_keys()` and wrap them with a `HirarchyHandler` 

By ordering the keys list we make sure the Handlers get grouped in a HirarchyHandler and called in the correct sequence such that the first handler which handles the event will suppress all following handlers.

Here is an Exampel:
 - We have 3 handlers for key "b":
   - `b` maps to "d"
   - `a+b` maps to "e"
   - `a+b+c` maps to "f"
 - the ordered list is
   - `[a+b+c, a+b, b]`
 - suppose we hold down key "a"
   - input-remapper injects whatever is mapped to "a"
 - and then we press key "b"
   - input-remapper notifyes all non hierarchic handlers which care about `b`
   - `HierarchyMapper` will walk trough the list and notify the handlers:
     - `a+b+c`: notify returns `False`  since c is not pressed
     - `a+b`: notify returns `True`
     - `b`: will be notified with `suppress = True` (let the handler know about the key-down but suppress its injection)

jonasBoss: That is the Idea

jonasBoss: Can you test out the combination handling? The new system currently only supports combination -> single key mappings. No macros or anything fancy yet.

There is now a different set of rules for combinations:
1. combinations trigger if at least all keys of the combination are pressed down
2. combinations release if **any** key of the combination is released
3. before triggering all forwarded keys which are part of the combination get released
4. we don't care about the order in which keys are pressed
**Overlapping Combinations:**
5. each key press can only ever trigger one combination
6. a key which completes multiple combinations will only trigger the longest combination
7. if there are multiple longest combinations:
     - take the sum of all keycodes for each combination
     - trigger the one which has the largest sum
     - this ensures we get always the same result otherwise we would be dependent on the order in which the mappings got saved to the preset.

especially with overlapping combinations I got a few times stuck keys on the forwarded device, this is dependent on the order in which keys get pressed and released. Hopefully it is fixed now.

I did not modify the preset file (yet) so the GUI still works

sezanzeb: mapping:

`1 + 2` -> `b`

1. 1 down -> 1 written
2. 2 down -> b written
3. up 1 and 2
4. 2 down -> b written
5. 2 up and down -> more bs written

When I trigger the combination again, it starts working again and "2" is written, like it should.

It happens randomly, maybe after half of all combinations

```
1633351 15:24:00.841503 SPAM keycode_mapper.py:459: ((1, 28, 0)) ------------------ forwarding release
1633351 15:24:01.593660 SPAM keycode_mapper.py:512: ((1, 2, 1)) ------------------- maps to macro (m(a,m(0,h())), keyboard)
1633351 15:24:01.604211 DEBUG keycode_mapper.py:321: macro_sending (1, 30, 1)
1633351 15:24:01.625029 DEBUG keycode_mapper.py:321: macro_sending (1, 11, 1)
1633351 15:24:01.713712 SPAM keycode_mapper.py:427: ((1, 2, 0)) ------------------- releasing macro
1633351 15:24:01.713826 SPAM keycode_mapper.py:440: ((1, 2, 0)) ------------------- releasing key
1633351 15:24:01.724231 DEBUG keycode_mapper.py:321: macro_sending (1, 11, 0)
1633351 15:24:01.745016 DEBUG keycode_mapper.py:321: macro_sending (1, 30, 0)
1633351 15:24:02.177615 SPAM keycode_mapper.py:539: ((1, 57, 1)) ------------------ forwarding
1633351 15:24:02.297753 SPAM keycode_mapper.py:459: ((1, 57, 0)) ------------------ forwarding release
1633351 15:24:02.457766 SPAM keycode_mapper.py:512: ((1, 2, 1)) ------------------- maps to macro (m(a,m(0,h())), keyboard)
1633351 15:24:02.468133 DEBUG keycode_mapper.py:321: macro_sending (1, 30, 1)
1633351 15:24:02.488848 DEBUG keycode_mapper.py:321: macro_sending (1, 11, 1)
1633351 15:24:02.657657 SPAM mapping_handler.py:133: Key((1, 2, 1), (1, 3, 1)) ----- maps to ([(1, 48, 1), 'keyboard'])
1633351 15:24:02.657786 DEBUG mapping_handler.py:87: injecting form new mapping handler: (1, 48, 1)
1633351 15:24:02.865689 SPAM mapping_handler.py:133: Key((1, 2, 1), (1, 3, 1)) ----- maps to ([(1, 48, 0), 'keyboard'])
1633351 15:24:02.865804 DEBUG mapping_handler.py:87: injecting form new mapping handler: (1, 48, 0)
1633351 15:24:02.866097 SPAM keycode_mapper.py:469: ((1, 3, 0)) ------------------- unexpected key up
1633351 15:24:03.121736 SPAM keycode_mapper.py:389: ((1, 2, 1), (1, 57, 1)) ------- unknown combination
1633351 15:24:03.121839 SPAM keycode_mapper.py:539: ((1, 57, 1)) ------------------ forwarding
1633351 15:24:03.217768 SPAM keycode_mapper.py:459: ((1, 57, 0)) ------------------ forwarding release
1633351 15:24:03.305816 SPAM keycode_mapper.py:531: ((1, 2, 1), (1, 3, 1)) -------- maps to (48, keyboard)
1633351 15:24:03.401880 SPAM keycode_mapper.py:446: ((1, 3, 0)) ------------------- releasing (48, keyboard)
1633351 15:24:03.521923 SPAM keycode_mapper.py:531: ((1, 2, 1), (1, 3, 1)) -------- maps to (48, keyboard)
1633351 15:24:03.617858 SPAM keycode_mapper.py:446: ((1, 3, 0)) ------------------- releasing (48, keyboard)
1633351 15:24:03.705938 SPAM keycode_mapper.py:531: ((1, 2, 1), (1, 3, 1)) -------- maps to (48, keyboard)
1633351 15:24:03.849909 SPAM keycode_mapper.py:446: ((1, 3, 0)) ------------------- releasing (48, keyboard)
```

sezanzeb: At the bottom of the log it should say

```
1665743 15:33:37.631813 SPAM keycode_mapper.py:539: ((1, 3, 1)) ------------------- forwarding
1665743 15:33:37.751806 SPAM keycode_mapper.py:459: ((1, 3, 0)) ------------------- forwarding release
```

instead

jonasBoss: oooh that is unfortunate. The old keycode-mapper is doing stuff which is messing everything up. I should have expected that. Will remove it from the consumer-control

jonasBoss: now it should work.

sezanzeb: `1 + 2` -> `a`
`2 + 1` -> `b`

```
2060570 17:00:20.303041 SPAM mapping.py:217: Key((1, 2, 1), (1, 3, 1)) maps to ('a', 'keyboard')
2060570 17:00:20.303098 SPAM mapping.py:217: Key((1, 3, 1), (1, 2, 1)) maps to ('b', 'keyboard')
```

pressing 1 and 2 prints:

```
2060709 17:01:39.746127 SPAM mapping_handler.py:133: Key((1, 2, 1), (1, 3, 1)) ----- maps to ([(1, 30, 1), 'keyboard'])
2060709 17:01:39.746269 DEBUG mapping_handler.py:87: injecting form new mapping handler: (1, 30, 1)
2060709 17:01:39.922219 SPAM mapping_handler.py:133: Key((1, 3, 1), (1, 2, 1)) ----- maps to ([(1, 48, 0), 'keyboard'])
2060709 17:01:39.922365 DEBUG mapping_handler.py:87: injecting form new mapping handler: (1, 48, 0)
2060709 17:01:39.922639 SPAM mapping_handler.py:133: Key((1, 2, 1), (1, 3, 1)) ----- maps to ([(1, 30, 0), 'keyboard'])
2060709 17:01:39.922712 DEBUG mapping_handler.py:87: injecting form new mapping handler: (1, 30, 0)

Comment (jonasBoss) [reply inputremapper/injection/context.py]: So far this is by design. But a good point, I should change it. And I need to find a way to supress all injection by the mapping which did not trigger.

Comment (jonasBoss) [reply inputremapper/injection/context.py]: okay, now `1+2` will actually write a, and `2+1` will write b
Also we no longer send unnecessary release events to the mapped preset.

there are still undefined edge cases like:
`1+2` -> `a`
`3+2` -> `b`
what will happen when pressing `1+3+2`? Either `a` or `b` will be sent, this is currently undefined behavior. Most likely `b` should be triggered. Resolving this would require us to keep track of the order of all currently pressed keys, and even then I am not sure if it is possible with my current approach.

Comment (sezanzeb) [reply inputremapper/injection/context.py]: I also think that `b` would be correct. But no strong feelings on my side.

What should work is that if only `1 + 2 + 3` -> `a` is mapped, and `0 + 2 + 1 + 3` is typed, the result should be "021a". Everything else is probably also undefined in the current behavior

Comment (jonasBoss) [reply inputremapper/injection/context.py]: > What should work is that if only 1 + 2 + 3 -> a is mapped, and 0 + 2 + 1 + 3 is typed, the result should be "021a". Everything else is probably also undefined in the current behavior

that is the case with the new injection. 

I am quiet happy with the combination handling as it is now. I would leave it like that if there are no objections.

File: inputremapper/injection/macros/macro.py [hunk 2]
```diff
@@ -209,40 +213,7 @@ def __init__(self, code, context):
...
-        """
-        while True:
-            await self._new_event_arrived.wait()
-            self._new_event_arrived.clear()
-
-            if filter is not None:
-                if not filter(self._newest_event, self._newest_action):
-                    continue
-
-            break
```
jonasBoss, line 245: There is a whole lot of stuff gone from the macros now. I didn't feel like notifying each macro about every event and then also waiting for a few asyncio Iterations. 

so I rewrote the `add_if_single()` function. Now the context keeps track of the `last_btn_down` and `last_btn_up` event.
`if_single` now waits until the release event or the timeout and compares the `triggering_event` with the `last_btn_down` from the context.

Can you please test with a few macros, especially the `if_single` function? It should be all the same except the `if_single` function now injects after the release or timeout.

sezanzeb: I'll do later

inputremapper/injection/consumer_control.py
```diff
@@ -93,17 +103,32 @@ async def run(self):
...
                 continue
 
+            tasks = []
+            results = []
+            if (event.type, event.code) in self.context.callbacks.keys():
+                for callback in self.context.callbacks[(event.type, event.code)]:
+                    tasks.append(callback(copy_event(event),
+                                          source=self._source,
+                                          forward=self._forward_to
+                                          ))
```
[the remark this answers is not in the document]
jonasBoss: you are right, that is confusing. I will untangle that mess.

inputremapper/injection/context.py
```diff
@@ -84,6 +164,72 @@ def update_purposes(self):
         self.left_purpose = self.mapping.get("gamepad.joystick.left_purpose")
         self.right_purpose = self.mapping.get("gamepad.joystick.right_purpose")
 
+    def update_callbacks(self) -> None:
+        """add the notify method from all sorted_handlers to self.callbacks"""
+        for key, handler_list in self._sorted_handlers.items():
+            self.callbacks[key[:2]] = []
+            for handler in handler_list:
+                self.callbacks[key[:2]].append(handler.notify)
```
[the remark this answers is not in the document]
sezanzeb: is it mapping (type, code) to the callback?

Gamepad D-Pads report something like `(EV_ABS, ABS_HAT0X, 1)` and `(EV_ABS, ABS_HAT0X, -1)`, for each direction. The release event for both of them has the value of 0.

inputremapper/injection/consumers/mapping_handler.py
```diff
@@ -0,0 +1,379 @@
...
+        ...
+
+    async def run(self) -> None:
+        ...
+
+
+class CombinationHandler:
+    """keeps track of a combination and notifies a sub handler
+
+    adheres to the MappingHandler protocol
```
sezanzeb, line 98: why is this not inheriting from `MappingHandler`?

inputremapper/injection/context.py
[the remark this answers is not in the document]
sezanzeb: So we could also call `.notify` on all Handlers and they will figure out themselves if they are going to do anything?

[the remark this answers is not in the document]
jonasBoss: exactly. That would create some unnecessary overhead, but it would work.

inputremapper/injection/consumers/mapping_handler.py
jonasBoss: That was my first approach, but it turns out that the MappingHandler as a super class is not very useful. I ended up overriding pretty much all of its methods with every handler. It didn't implement any useful logic. 

I made the MappingHandler a Protocol class: https://www.python.org/dev/peps/pep-0544/
It serves as a useful type-hint, in the sens that any object which has the in `MappingHandler` specified properties and methods is considered to be a `MappingHandler` by most IDEs and type-checking.

I should probably rename it to something like `TopLevelHandler`, to make clear that this handler might be directly exposed to the context, and therefor must implement the methods and properties as described by the protocol.

A handler like the `KeyHandler` or `MacroHandler` does not follow the `MappingHandler` protocol if we try to add such a handler to the `context._handlers` dict we will get a type warning by the IDE

inputremapper/injection/macros/macro.py
sezanzeb: > except the if_single function now injects after the release

yeah, I just wanted to point that out. Other than that it seems to work fine

inputremapper/injection/consumers/mapping_parser.py
```diff
@@ -0,0 +1,199 @@
...
+        if sub_key[2] == 1:
+            sub_keys.append((*sub_key[:2], 10))  # trigger point at 10%
+        else:
+            assert sub_key[2] == -1
+            sub_keys.append((*sub_key[:2], -10))  # trigger point at -10%
+    new_key = Key(*sub_keys)
+
+    config = {
+        "key": new_key,
+        "symbol": symbol_and_target[0],
```
sezanzeb, line 191:
```suggestion
        "symbol": clean(symbol_and_target[0]),
```

inputremapper/injection/macros/macro.py
jonasBoss: I am not sure if can do anything about that. Do you think this is a huge issue? 
Considering that it only effects the execution of the `other` macro and only if it is not executed by the timeout.

inputremapper/injection/consumers/mapping_parser.py
jonasBoss: on that note: I just inflated the log a bit more.
I created a custom logger which parses a MappingHandler object and spits out the path a event takes through the different handlers. It helps a lot for debugging and understanding the handler interaction:
image
creates this log now:
```
DEBUG mapping_parser.py:75: created mapping handler:
DEBUG mapping_parser.py:76: HierarchyHandler for (1, 310) <140107968623808>:
DEBUG mapping_parser.py:76:     CombinationHandler for ((1, 310, 1), (3, 3, 10)) <140107968622608>:
DEBUG mapping_parser.py:76:         KeyHandler <140107968622800>:
DEBUG mapping_parser.py:76:             maps to: (1, 2) on keyboard
DEBUG mapping_parser.py:76:     CombinationHandler for ((1, 310, 1),) <140107968623424>:
DEBUG mapping_parser.py:76:         KeyHandler <140107968623616>:
DEBUG mapping_parser.py:76:             maps to: (1, 4) on keyboard
DEBUG mapping_parser.py:81: created mapping handler:
DEBUG mapping_parser.py:82: AbsToBtnHandler for (3, 3, 10) <140107968623712>:
DEBUG mapping_parser.py:82:     HierarchyHandler for (3, 3) <140107968624048>:
DEBUG mapping_parser.py:82:         CombinationHandler for ((1, 310, 1), (3, 3, 10)) <140107968622608>:
DEBUG mapping_parser.py:82:             KeyHandler <140107968622800>:
DEBUG mapping_parser.py:82:                 maps to: (1, 2) on keyboard
DEBUG mapping_parser.py:82:         CombinationHandler for ((3, 3, 10),) <140107968622992>:
DEBUG mapping_parser.py:82:             KeyHandler <140107968623184>:
DEBUG mapping_parser.py:82:                 maps to: (1, 3) on keyboard
```

inputremapper/injection/macros/macro.py
sezanzeb: Some people expect this behavior (https://github.com/sezanzeb/input-remapper/issues/183). `if_single` was specifically designed for this use-case.

Maybe you can move `_wait_for_event` to the context, so that `if_single` can use it like it used to

sezanzeb: There might also be a use case for a macro that waits for a specific key in the future. Like `wait_for(KEY_A)` or something. I think it would be good if it would be theoretically possible to more or less easily add something like this.

If macros can listen for events (like `.connect` in gtk, `addEventListener` in javascript or ) it might result in a design that works well for this.

That reminds me, I think connecting to events is better design than passing callback functions. Maybe something like https://www.joeltok.com/blog/2021-3/building-an-event-bus-in-python

jonasBoss: I reworked the if_single function again. Please have a look at it. It should now work as intended.

jonasBoss: well ...
Except for some race conditions. I Guess this is why there is a small delay in the keycode_mapper

inputremapper/injection/consumer_control.py
```diff
@@ -93,23 +104,40 @@ async def run(self):
                 # won't appear, no need to forward or map them.
                 continue
 
+            for listener in self.context.listeners:
+                asyncio.create_task(listener(event))
+                # allow the just created task to do some stuff
+                await asyncio.sleep(0)
```
sezanzeb, line 110: can you reuse the complete old `delayed_handle_keycode` function here? Especially so that the docstring with the explanation is preserved

sezanzeb, line 110:
```suggestion
            for listener in self.context.listeners:
                asyncio.create_task(listener(event))
                
            # give macros a priority of working on their asyncio iterations
            # first before handle_keycode. This is important for if_single.
            # If if_single injects a modifier to modify the key that canceled
            # its sleep, it needs to inject it before handle_keycode injects
            # anything. This is important for the space cadet shift.
            # 1. key arrives
            # 2. stop if_single
            # 3. make if_single inject `then`
            # 4. inject key
            # But I can't just wait for if_single to do its thing because it might
            # be a macro that sleeps for a few seconds.
            # This appears to me to be incredibly race-conditiony. For that
            # reason wait a few more asyncio ticks before continuing.
            # But a single one also worked. I can't wait for the specific
            # macro task here because it might block forever. I'll just give
            # it a few asyncio iterations advance before continuing here.
            for _ in range(10):
                # Noticable delays caused by this start at 10000 iterations
                # Also see the python docs on asyncio.sleep. Sleeping for 0
                # seconds just iterates the loop once.
                await asyncio.sleep(0)
```

sezanzeb, line 110: According to the docstring in the old `delayed_handle_keycode` function this was, so that if_single gets a chance to first start its then/else child-macro before the key is used to do the next macro.

If `if_single` injects a modifier, this modifier should be active before the next handler injects an "a" or something, so that it is possible to capitalize it via `if_single`.

1. event from keyboard arrives
2. the asyncio.event if_single is waiting for triggers
3. if_single decides runs then
4. the original event is forwarded (or whatever it is supposed to do)

sezanzeb: The listeners are called before he handlers inject/forward anything, so your changes are probably correct

```suggestion
            for listener in self.context.listeners:
                asyncio.create_task(listener(event))
                
            # Running macros have priority, give them a head-start for processing the
            # event.  If if_single injects a modifier, this modifier should be active
            # before the next handler injects an "a" or something, so that it is
            # possible to capitalize it via if_single.
            # 1. Event from keyboard arrives (e.g. an "a")
            # 2. The asyncio.event if_single is waiting for triggers
            # 3. if_single decides runs then (e.g. injects shift_L)
            # 4. The original event is forwarded (or whatever it is supposed to do)
            # 5. Capitalized "A" is injected.
            # So make sure to call the listeners before notifying the handlers.
            await asyncio.sleep(0)
```

inputremapper/injection/context.py
```diff
@@ -84,6 +110,14 @@ def update_purposes(self):
         self.left_purpose = self.mapping.get("gamepad.joystick.left_purpose")
         self.right_purpose = self.mapping.get("gamepad.joystick.right_purpose")
 
+    def create_callbacks(self) -> None:
+        """add the notify method from all _handlers to self.callbacks"""
+        for key, handler_list in self._handlers.items():
+            if key[0][:2] not in self.callbacks.keys():
+                self.callbacks[key[0][:2]] = []
+            for handler in handler_list:
+                self.callbacks[key[0][:2]].append(handler.notify)
```
sezanzeb, line 119: could this also use `.listeners`?

how would it be possible to ensure the listeners for macros run first? Via a priority parameter?

inputremapper/injection/consumer_control.py
```diff
@@ -93,23 +104,40 @@ async def run(self):
...
+                        ev, source=self._source, forward=self._forward_to
+                    )
+                    tasks.append(coroutine)
+                results = await asyncio.gather(*tasks)
+
+            if True in results:
+                continue
+
+            # try legacy injection if the new injection did not do anything
+            # TODO: Remove. only keep the call to forward
```
sezanzeb, line 127: By using pythons for/break/else logic the boolean handled can be avoided:

```py
            for consumer in self._consumers:
                # copy so that the consumer doesn't screw this up for
                # all other future consumers
                event_copy = copy_event(event)
                if consumer.is_handled(event_copy):
                    await consumer.notify(event_copy)
                    break
            else:
                # forward the rest
                if event.type == evdev.ecodes.EV_KEY:
                    logger.debug_key((event.type, event.code, event.value), "forwarding")
                self._forward_to.write(event.type, event.code, event.value)
                # this already includes SYN events, so need to syn here again
```

inputremapper/injection/consumers/mapping_handler.py
```diff
@@ -504,3 +497,163 @@ async def notify(
...
+
+        output_value = self._calc_qubic(input_value, self._expo)
+        self._last_value = output_value * scale_factor * self._gain
+
+        if not self._running:
+            asyncio.ensure_future(self._run())
+        return True
+
+    @staticmethod
+    def _calc_qubic(x: float, k: float) -> float:
```
sezanzeb, line 578: is this for making the cursor move faster when the abs value is at its maximum?

inputremapper/injection/consumers/mapping_parser.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+def _create_hierarchy_handlers(handlers: Dict[Key, MappingHandler]) -> Dict[Key, MappingHandler]:
+    """sort handlers by sub_keys and create Hierarchy handlers"""
+    # gather all single keys in all mappings
+    sorted_handlers = {}
+    all_keys = handlers.keys()
+    keys = set()  # set of keys with len 1
+    for key in all_keys:
+        for sub_key in key:
```
sezanzeb, line 119: do you have an idea for a better name for the `Key` class? Since it contains one or more keys the name is not very fitting.

I also don't like the term `sub_key`, but I just couldn't come up with something better back then.

How about there is a `class Combination` that contains multiple of `class Key` instead? By doing so, one could write `combination[0].value`, which is much easier to understand than `key[0][0]`

inputremapper/injection/consumers/mapping_parser.py
```diff
@@ -0,0 +1,212 @@
...
+
+MappingHandlers = Dict[Key, List[MappingHandler]]
+
+mapping_handler_classes: Dict[str, Type[MappingHandler]] = {
+    # all available mapping_handlers
+    "combination": CombinationHandler,
+}
+
+
+def parse_mapping(mapping: Mapping, context: ContextProtocol) -> MappingHandlers:
```
sezanzeb, line 46: I think the `Mapping` class should be renamed to `Preset`

inputremapper/injection/consumers/mapping_parser.py
```diff
@@ -0,0 +1,212 @@
...
+    """create dict with a list of MappingHandlers for each key
+
+    Key is of len 1
+    """
+
+    # create a handler for each mapping
+    combination_handlers = {}
+    normal_handlers = {}
+
+    for key, sub_mapping in mapping:
```
sezanzeb, line 56: And maybe we should change the iterator of the `Mapping` class to yield `key, symbol, target`

inputremapper/injection/macros/macro.py
sezanzeb:
```c#
if_single(
  key(a),
  key(b)
)
```

never releases the b. It says "Injector coroutines ended"

```
16:43:14.903511 3299005 Service INFO config.py:247: Loaded config from "/home/mango/.config/input-remapper/config.json"
16:43:14.904232 3299005 Service DEBUG daemon.py:267: Refreshing because last info is too old
16:43:15.005099 3299005 Service DEBUG groups.py:336: Discovering device paths
16:43:15.005685 3299005 Service DEBUG groups.py:377: Found "3_1_1_input-remapper", "/dev/input/event15", "input-remapper mouse", type: mouse
16:43:15.006021 3299005 Service DEBUG groups.py:377: Found "3_1_1_input-remapper", "/dev/input/event14", "input-remapper gamepad", type: gamepad
16:43:15.006624 3299005 Service DEBUG groups.py:377: Found "3_1_1_input-remapper", "/dev/input/event13", "input-remapper keyboard", type: keyboard
16:43:15.006915 3299005 Service DEBUG groups.py:377: Found "3_7119_5_usb-0000:28:00.3-4.3", "/dev/input/event5", "USB Optical Mouse", type: mouse
16:43:15.007219 3299005 Service DEBUG groups.py:377: Found "3_1118_2048_usb-0000:28:00.3-4.1", "/dev/input/event12", "Microsoft Microsoft® Nano Transceiver v2.0 System Control", type: unknown
16:43:15.007691 3299005 Service DEBUG groups.py:377: Found "3_1118_2048_usb-0000:28:00.3-4.1", "/dev/input/event11", "Microsoft Microsoft® Nano Transceiver v2.0 Consumer Control", type: keyboard
16:43:15.008119 3299005 Service DEBUG groups.py:377: Found "3_1118_2048_usb-0000:28:00.3-4.1", "/dev/input/event9", "Microsoft Microsoft® Nano Transceiver v2.0 Mouse", type: mouse
16:43:15.008493 3299005 Service DEBUG groups.py:377: Found "3_1118_2048_usb-0000:28:00.3-4.1", "/dev/input/event8", "Microsoft Microsoft® Nano Transceiver v2.0", type: keyboard
16:43:15.009048 3299005 Service DEBUG groups.py:377: Found "3_1133_49960_usb-0000:03:00.0-1", "/dev/input/event4", "Logitech USB Keyboard System Control", type: unknown
16:43:15.009433 3299005 Service DEBUG groups.py:377: Found "3_1133_49960_usb-0000:03:00.0-1", "/dev/input/event3", "Logitech USB Keyboard Consumer Control", type: unknown
16:43:15.009835 3299005 Service DEBUG groups.py:377: Found "3_1133_49960_usb-0000:03:00.0-1", "/dev/input/event2", "Logitech USB Keyboard", type: keyboard
16:43:15.010552 3299005 Service INFO groups.py:448: Found "input-remapper mouse", "USB Optical Mouse", "Microsoft Microsoft® Nano Transceiver v2.0", "Logitech USB Keyboard"
16:43:15.010805 3299005 Service INFO mapping.py:179: Loading preset from "/home/mango/.config/input-remapper/presets/Logitech USB Keyboard/new preset 2.json"
16:43:15.011006 3299005 Service DEBUG mapping.py:218: Key((1, 2, 1),) maps to ('if_single(\n  key(a),\n  key(b)\n)', 'keyboard')
16:43:15.014825 3299005 Service INFO injector.py:161: Stopping injecting keycodes for group "Logitech USB Keyboard"
16:43:15.038459 3299005 Service DEBUG daemon.py:461: Using keycodes from "/home/mango/.config/input-remapper/xmodmap.json"
16:43:15.038890 3299005 Service DEBUG system_mapping.py:144: Updated keycodes with 0 new ones
16:43:15.041251 3299704 Service INFO injector.py:300: Starting injecting the mapping for "Logitech USB Keyboard"
16:43:15.042482 3299704 Service DEBUG parse.py:416: preparing macro if_single(key(a),key(b)) for later execution
16:43:15.042841 3299704 Service DEBUG parse.py:252: calls if_single with key(a),key(b)
16:43:15.043173 3299704 Service DEBUG parse.py:252:   calls key with a
16:43:15.043293 3299704 Service DEBUG parse.py:309:     string a
16:43:15.043381 3299704 Service DEBUG parse.py:273:   add call to key with ['a'], {}
16:43:15.043713 3299704 Service DEBUG parse.py:252:   calls key with b
16:43:15.043823 3299704 Service DEBUG parse.py:309:     string b
16:43:15.043910 3299704 Service DEBUG parse.py:273:   add call to key with ['b'], {}
16:43:15.044044 3299704 Service DEBUG parse.py:273: add call to if_single with [<Macro "key(a)">, <Macro "key(b)">], {}
16:43:15.044249 3299704 Service DEBUG mapping_parser.py:76: created mapping handler:
16:43:15.044401 3299704 Service DEBUG mapping_parser.py:77: CombinationHandler for ((1, 2, 1),) <140576378363376>:
16:43:15.044489 3299704 Service DEBUG mapping_parser.py:77:     MacroHandler <140576378361552>:
16:43:15.044585 3299704 Service DEBUG mapping_parser.py:77:         maps to <Macro "if_single(key(a),key(b))"> on keyboard
16:43:15.044819 3299704 Service DEBUG injector.py:210: No need to grab /dev/input/event4
16:43:15.045255 3299704 Service DEBUG injector.py:210: No need to grab /dev/input/event3
16:43:15.045570 3299704 Service DEBUG injector.py:197: Grabbing "/dev/input/event2" because of "Key((1, 2, 1),)"
16:43:15.045749 3299704 Service DEBUG injector.py:217: Grab /dev/input/event2
16:43:15.179320 3299704 Service DEBUG consumer_control.py:94: Starting to listen for events from /dev/input/event2, fd 44
16:43:16.886627 3299704 Service DEBUG mapping_handler.py:160: Key((1, 2, 1)) ---------------- triggered: sending to sub-handler
16:43:17.190412 3299704 Service DEBUG mapping_handler.py:277: (1, 48, 1) -------------------- sending from macro to keyboard
16:43:17.190733 3299704 Service DEBUG injector.py:361: Injector coroutines ended
```

inputremapper/injection/consumers/mapping_handler.py
jonasBoss: Yes, have a look at this:
https://www.geogebra.org/calculator/mkdqueky

sezanzeb: cool, can we have this link in the docstring?

inputremapper/injection/consumer_control.py
jonasBoss: good idea. But I still want it to be inside the for loop so that we give more time if there are more listeners, and don't sleep if there is nor reason to sleep

inputremapper/injection/consumers/mapping_handler.py
jonasBoss: I am not sure how long lived that is. But sure I will add it

inputremapper/injection/consumers/mapping_parser.py
jonasBoss: Yeah it is currently super confusing I like the idea of using two different classes. The new injection actually depends on the fact that there is a combination, which contains one or more keys, and a key which only has len 1.
I would avoid `Key`, it is way to ambiguous and clashes with python method names like `dict.keys()`

What about `EventCombination` and `EventTuple`

inputremapper/injection/consumers/mapping_parser.py
jonasBoss: I was thinking something like this as well. A preset is a collection of mappings, so renaming this would fit well.

inputremapper/injection/consumers/mapping_parser.py
sezanzeb: I think there would be a bit of an ambiguity between `EventTuple` and `evdev.InputEvent`.

I'd use `evdev.InputEvent` inside `EventCombination`, but it requires a timestamp to be set.

```py
class EventCombination:
    def add_event(type, code, value, target):
        self._events.append(evdev.InputEvent(
            sec=0,
            usec=0,
            type=type,
            code=code,
            value=value
        ))

    def __hash__(self):
        return hash(((event.type, event.code, event.value) for event in self._events))
```

If we never use the evdev.InputEvent constructor directly, this is not an issue. We can just insert dummy numbers for the timestamp.

inputremapper/injection/consumers/mapping_parser.py
jonasBoss: let us rename the mapping to preset, as you suggested.
And then I think the preset should yield a mapping object. Which would be a Dataclass
I would like to do something like:
```
for mapping in preset:
   handlers[mapping.key] = MappingHandler(mapping)
```
A mapping can be so much more than just some keys to other keys or macros. For each abs axis we can individually specify the gain, expo, deadzone, ...
We also can map multiple buttons to one ABS axis by just specifying  different trigger_points.
currently I have a lot of methods in `mapping_parser`  which construct different dicts for the mappings, but all that should happen in the preset class an ultimately represented in the save file

inputremapper/injection/consumers/mapping_parser.py
jonasBoss: No, they would not work:
```
>> e1 = evdev.InputEvent(0,0,1,5,1)
>> e2 = evdev.InputEvent(0,0,1,5,1)
>> e1 == e2
False
>> e1.__hash__() == e2.__hash__()
False
```
we need both of these properties to be True

inputremapper/injection/macros/macro.py
jonasBoss: that is new, now I have to go back and see when I broke it

inputremapper/injection/consumers/mapping_parser.py
sezanzeb: but

```py
event_combination_1 = EventCombination()
event_combination_1.add_event(1, 5, 1)

event_combination_2 = EventCombination()
event_combination_2.add_event(1, 5, 1)

hash(event_combination_1) == hash(event_combination_2)
```

would

inputremapper/injection/injector.py
```diff
@@ -348,9 +348,6 @@ def run(self) -> None:
 
         try:
             loop.run_until_complete(asyncio.gather(*coroutines))
-        except RuntimeError:
-            # stopped event loop most likely
-            pass
```
jonasBoss, line 342: Was there a good reason to have this?

sezanzeb: might be `quick_cleanup`.

Also, there happens to be another `except RuntimeError` in `quick_cleanup`, with the explanation `# happens when the event loop disappears for magical reasons`. So I probably never solved the mystery why this happens, and I think it only happens in tests.

Why are you removing it?

jonasBoss: Nevermind, I know now why Unfortunately it catches to many errors. we should probably at least print something to the debug log

sezanzeb: I think it might be important to catch exceptions here in order to ungrab the devices properly before finishing with `run`.

Consider something like

```py
        except RuntimeError as error:
            # stopped event loop most likely
            logger.error(error)
            pass
```

inputremapper/injection/macros/macro.py
jonasBoss: should be fixed now

inputremapper/injection/injector.py
```diff
@@ -337,9 +348,9 @@ def run(self):
...
+        except RuntimeError as error:
+            if error.__str__() != "Event loop stopped before Future completed.":
+                raise error
```
sezanzeb, line 353:
```suggestion
        except RuntimeError as error:
            if str(error) != "Event loop stopped before Future completed.":
                raise error
            logger.debug(error)
```

inputremapper/injection/injector.py
jonasBoss: stopping the event loop from the _msg_listener() causes the exception.
maybe there is a way to cancel all futures when the _msg_listener() receives CLOSE, then it should stop automatically.
for now I do it like this:
```
 try:
    loop.run_until_complete(asyncio.gather(*coroutines))
except RuntimeError as error:
    if error.__str__() != "Event loop stopped before Future completed.":
        raise error
except OSError as error:
    logger.error("Failed to run injector coroutines: %s", str(error))
```

sezanzeb:
```py
 try:
    loop.run_until_complete(asyncio.gather(*coroutines))
except RuntimeError as error:
    # the loop might have been stopped via a `CLOSE` message,
    # which causes the error message below. This is expected behavior
    if str(error) != "Event loop stopped before Future completed.":
        raise error
except OSError as error:
    logger.error("Failed to run injector coroutines: %s", str(error))
```

inputremapper/injection/consumers/mapping_parser.py
jonasBoss: No, I currently use a Key of length 1 for all handlers in `context._handlers` and `context.callbacks`.
But this is confusing because it is not clear that we cannot have a combination there. I would rather use some identifier that is distinct from the Identifier which can be a combination.

jonasBoss: I think a dataclass should do everything we need for us

```
from dataclasses import dataclass
>>> @dataclass
... class EventIdentifier:
...     type: int
...     code: int
...     value: int
    
>>> e1 = EventIdentifier(1,5,1)
>>> e2 = EventIdentifier(1,5,1)
>>> e1 == e2
True
>>> e1.__hash__ == e2.__hash__
True
```

sezanzeb: > it is not clear that we cannot have a combination there

I see

Or we patch evdev.InputEvent to support __eq__

```py
>>> import evdev
>>> from evdev.ecodes import EV_KEY, KEY_A
>>> 
>>> def __eq__(self, other):
...     return (self.type, self.code, self.value) == (other.type, other.code, other.value)
... 
>>> evdev.InputEvent.__eq__ = __eq__
>>> 
>>> print(hash(evdev.InputEvent(0, 0, EV_KEY, KEY_A, 1)))
8742126302377
>>> print(hash(evdev.InputEvent(0, 0, EV_KEY, KEY_A, 1)))
8742126302377
>>> print(evdev.InputEvent(0, 0, EV_KEY, KEY_A, 1) == evdev.InputEvent(0, 0, EV_KEY, KEY_A, 1))
True
>>> print(evdev.InputEvent(0, 0, EV_KEY, KEY_A, 1) == evdev.InputEvent(0, 0, EV_KEY, KEY_A, 0))
False
>>> 
```

sezanzeb: Ah no we also need to patch `__hash__`, otherwise all InputEvent objects have the same hash:

```py
#!/usr/bin/python

import evdev
from evdev.ecodes import EV_KEY, KEY_A

def __eq__(self, other):
    return (self.type, self.code, self.value) == (other.type, other.code, other.value)

def __hash__(self):
    return hash((self.type, self.code, self.value))

evdev.InputEvent.__eq__ = __eq__
evdev.InputEvent.__hash__ = __hash__

print(hash(evdev.InputEvent(1000, 1000, EV_KEY, KEY_A, 1)))
print(hash(evdev.InputEvent(2000, 3000, EV_KEY, KEY_A, 1)))
print(hash(evdev.InputEvent(0, 0, EV_KEY, KEY_A, 2)))
print(evdev.InputEvent(0, 0, EV_KEY, KEY_A, 1) == evdev.InputEvent(0, 0, EV_KEY, KEY_A, 1))
print(evdev.InputEvent(0, 0, EV_KEY, KEY_A, 1) == evdev.InputEvent(0, 0, EV_KEY, KEY_A, 0))
```

```
print(hash(evdev.InputEvent(1000, 1000, EV_KEY, KEY_A, 1)))
print(hash(evdev.InputEvent(2000, 3000, EV_KEY, KEY_A, 1)))
print(hash(evdev.InputEvent(0, 0, EV_KEY, KEY_A, 2)))
print(evdev.InputEvent(0, 0, EV_KEY, KEY_A, 1) == evdev.InputEvent(0, 0, EV_KEY, KEY_A, 1))
print(evdev.InputEvent(0, 0, EV_KEY, KEY_A, 1) == evdev.InputEvent(0, 0, EV_KEY, KEY_A, 0))

If we use evdev.InputEvent, indexing datastructures with incoming events like

```
something = {
    evdev.InputEvent(0, 0, EV_KEY, KEY_A, 1): "foo"
}

event = device.read_one()
print(something[event])  # "foo"
```

would be possible

Comment (jonasBoss) [reply inputremapper/injection/consumers/mapping_parser.py:140]: I don't like the idea of patching evdev or any other module. We don't know how evdev works internally. It feels a bit error prone. Also it is confusing if InputEvent does not behave how it is documented in the evdev documentation.

I would rather create a new class. That way we have full control over the behavior we can give it some useful functions like:
```
@dataclass
class EventData:
    type: int
    code: int
    value: int
    
    def match_input_event(self, event: evdev.InputEvent) -> bool:
       ...

    @classmethod
    def create_from_event(cls, event: evdev.InputEven) -> EventData:
        ...
```

Comment (sezanzeb) [reply inputremapper/injection/consumers/mapping_parser.py:140]: I think seems really confusing that there would be an object that holds the exact same information as an `evdev.InputEvent`, has "Event" in the name, but is not actually an event and something different from `evdev.InputEvent`. But I can't come up with good alternatives.

What about converting the `evdev.InputEvent` into our own class (which inherits from `evdev.InputEvent`) after reading? Whenever we see code that goes like `foo.value` we know that this must be an `inputremapper.InputEvent` object. It would have `__hash__` and `__eq__`, we are not patching evdev, and there is no confusion between two very similar classes. Downside is that we have to remember to use that one instead of the one from python-evdev.

Comment (sezanzeb) [reply inputremapper/injection/consumers/mapping_parser.py:140]: we are already wrapping evdev.UInput, so the idea is not entirely new

Comment (jonasBoss) [reply inputremapper/injection/consumers/mapping_parser.py:140]: Replacing all events with our own version seems like a good idea. We would convert them directly in the `async_read_loop()` and can override different functionality without affecting evdev.

I don't think we need to inherit from `evdev.InputEvent`, since that does not provide any useful functionality.

I would even argue that we should not inherit from `evdev.InputEvent`, instead create it with `@dataclass(frozen=True)` that way we can not change a object after creation and we never need to worry if we should copy a event before we pass it to a function.

Comment (sezanzeb) [reply inputremapper/injection/consumers/mapping_parser.py:140]: great! Lets do it

File: inputremapper/injection/consumers/mapping_parser.py (hunk 13 above)
Comment (sezanzeb) [reply inputremapper/injection/consumers/mapping_parser.py:70]: Sounds intuitive, yes

**Comment by sezanzeb** [comment]
If macros fail to parse (due to using an unknown function name), it prints

```
  File "/usr/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/mnt/data/Code/input-remapper/inputremapper/injection/injector.py", line 350, in run
    loop.run_until_complete(asyncio.gather(*coroutines))
  File "/usr/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/mnt/data/Code/input-remapper/inputremapper/injection/consumer_control.py", line 125, in run
    results = await asyncio.gather(*tasks)
  File "/mnt/data/Code/input-remapper/inputremapper/injection/consumers/mapping_handler.py", line 161, in notify
    return await self._sub_handler.notify(ev)
  File "/mnt/data/Code/input-remapper/inputremapper/injection/consumers/mapping_handler.py", line 271, in notify
    self._macro.press_trigger(event)
AttributeError: 'NoneType' object has no attribute 'press_trigger'
```

File: inputremapper/injection/consumers/mapping_parser.py (hunk 13 above)
Comment (jonasBoss) [reply inputremapper/injection/consumers/mapping_parser.py]: I just renamed the Mapping to Preset, and went over all the docstrings, to untangle things. There might be some wrong docsrtings, which actually refer to a mapping but the word `mapping` got replaced with `preset`. I am not sure if I caught all of those.

Also there is a file `presets.py` which contains methods for managing presets. I will go over them and see if I can move them to the `Preset` class, make them member- or class-functions

File: inputremapper/gui/user_interface.py [hunk 16]
```diff
@@ -464,13 +465,18 @@ def check_macro_syntax(self):
...
 
-            error = parse(output, active_preset, return_errors=True)
-            if error is None:
+            error_sting = None
+            try:
+                parse(output, active_preset)
+            except MacroParsingError as error:
+                error_sting = f"{error.__class__.__name__}: {str(error)}"
+
+            if error_sting is None:
```
jonasBoss, line 474: I moved the exception catching out of the parse() function. And added a custom `MacroParsingError`. Now the creation of a `MacroHandler` will fail if the macro is not valid. 

Any `MappingHandler` that raises an `inputremapper.exceptions.Error` during initialization will not get added to the injection. 

TLDR: Invalid macros will now be ignored in the Injection process

jonasBoss:
Should I move all the different config related files to a common `configs` directory?
I think it would clean up the `inputremapper/` directory. So far my candidates for moving are:
 1. config.py
 2. data.py
 3. migrations.py
 4. paths.py
 5. preset.py
 6. system_mapping.py

sezanzeb:
ok, sounds good to me

inputremapper/configs/mapping.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+# TODO: in python 3.11 use enum.StrEnum
+class KnownUinput(str, enum.Enum):
+    keyboard = "keyboard"
+    mouse = "mouse"
+    gamepad = "gamepad"
+
+
+class Mapping(BaseModel):
```
jonasBoss, line 38: How do you feel about an added dependency: https://pydantic-docs.helpmanual.io/
I think it is a perfect fit for managing all configs which a user might change.

It provides automatic type casting and validation for variable assignment, and enforces the variable types defined in class definition.

I just implemented the Mapping class to see how it would work. I think the benefits are huge, and I could remove all the type and value validation inside the `MappingHandler`
```python
from pydantic import BaseModel, confloat
class Mapping(BaseModel):
    deadzone: confloat(ge=0, le=1) = 0.1

    class Config:
        validate_assignment = True

mapping = Mapping(deadzone="0.5")  # will automatically cast to float
mapping.deadzone = -1  # will raise an ValidationError: deadzone must be greater or equal to 0
```

sezanzeb: it seems to be available in both ubuntu and arch repos, sounds good to me

sezanzeb:
If a key-down event is injected into `keyboard`, and then the injection stopped, the key will be kept pressed forever. This is fatal if the key is the super key, which means the computer cannot be controlled at all anymore. Maybe the various handlers can make sure to release all injected keys when they stop, possibly in `__del__`

jonasBoss:
I replaced all internal evdev events with our new `InputEvent` class. The `Key` class is now a `EventCombination` class, which inherits directly from tuple. It is essentially a tuple of `InputEvent` objects. 

With all the renaming and moving stuff around, this gets kinda huge.
But at least the integration tests are still passing, except for one:
`test_gamepad_purpose_mouse_and_button` which seems to fail because:
`Exception: Ubuntus version of evdev doesn't support .absinfo` I rely on absinfo in some mapping handlers is there a alternative?

sezanzeb:
> Exception: Ubuntus version of evdev doesn't support .absinfo I rely on absinfo in some mapping handlers is there a alternative?

`device.capabilities(absinfo=True)`

jonasBoss:
Now we have access to all the new mappings that are possible. The new mapping is exposed to the `preset.json` and we have a migration function. Better **Backup your presets** before testing anything.

Now some fun stuff like this is possible:
```json
{
    "3,0,-10": {
        "target_uinput": "keyboard",
        "output_symbol": "a"
    },
    "3,0,-30": {
        "target_uinput": "keyboard",
        "output_symbol": "b"
    },
    "3,0,-50": {
        "target_uinput": "keyboard",
        "output_symbol": "c"
    },
    "3,0,0+3,0,50": {
        "target_uinput": "mouse",
        "output_type": 2,
        "output_code": 0,
        "gain": 0.43192002721555645
    },
    "3,0,80":{
        "target_uinput": "keyboard",
        "output_symbol": "modify(SHIFT_L, key(h)).key(i)"
    }
}
```
move the left stick to the left and we trigger:
 - `3,0,-10` -> `a` at 10% travel
 - `3,0,-30` -> `b` at 30% travel
 - `3,0,-50` -> `c` at 50%

move the left stick to the right:
 - `3,0,0+3,0,50` -> `REL_X` The mouse will move to the right at 50% stick travel
 - `3,0,80` -> `macro` The macro will greet you, while the mouse keeps moving

move the stick back to the middle:
 - `3,0,0+3,0,50` -> `REL_X` The mouse stops moving if the stick is less than 50% to the right

sezanzeb:
> `3,0,-30` -> `b` at 30% travel

if two abs event arrive with values of 15% and 35%, what will be triggered?

jonasBoss:
> if two abs events arrive with values of 15% and 35%, what will be triggered?

both the `3,0,-10 -> a` and `3,0,-30 -> b` mappings receive all the abs events for event code 0, independent from each other because their EventCombination (Key) is different. So at 35% travel both `a` and `b` are pressed at the same time. If we move the stick back to the middle first `b` and than `a` gets released.

sezanzeb:
The configuration file section needs to be updated: https://github.com/sezanzeb/input-remapper/blob/main/readme/usage.md#configuration-files

Should the "how it works" section be updated or just removed? https://github.com/sezanzeb/input-remapper/blob/main/readme/development.md#how-it-works

jonasBoss:
> is it worth maintaining the sections that describe the inner workings?

I think it is a good idea to keep that up to date. We don`t need details there, but a description of the overall architecture should be there.

We also need some UI changes. I looked into supporting the current UI with the drop-down menus on the left, for mapping left an right stick on a controller. But I think there is no good way to do that. The information whether the left stick should map to mouse, scroll-wheel or buttons is no longer available in the preset.
We could try and figure out if it is implicitly mapped, but that is hard and can brake any time there are manual changes to the preset.

I have some ideas, I will try to make some mockups for them. But I am open for any suggestions on that front.

jonasBoss:
This is my current Idea for the UI. We can display a user defined label and the combination as fallback in the `selection_label_listbox`. 

The new listbox shows the combination in detail. Obviously we need to display human readable names there, and for mapping a axis as part of the combination we can have a slider to set the trigger threshold.

On the new page could also be a drop-down menu with predefined settings such as `left joystick to mouse`

Any feedback and ideas for changes are very welcome.

image
image

sezanzeb:
image

I don't understand this at all

sezanzeb:
mockup with spacings:

image

I'd do it differently though:

image

image

jonasBoss:
> https://user-images.githubusercontent.com/28510156/152954613-8bdcaa04-211f-412d-9cdb-036038a447ad.png
> I don't understand this at all
> 
> Please try to design the ui so that new users are not overwhelmed.

On `main` we have the ability to map joystick axis to buttons in the drop-down on the left. 
Earlier I wrote this:

> I looked into supporting the current UI with the drop-down menus on the left, for mapping left an right stick on a controller. But I think there is no good way to do that. The information whether the left stick should map to mouse, scroll-wheel or buttons is no longer available in the preset.
We could try and figure out if it is implicitly mapped, but that is hard and can brake any time there are manual changes to the preset.

The Listbox is a Idea to still support this feature in the UI. Because editing the config is, as you said, not a good experience. I think i will implement its functionality and then we can decide if it should be visible or if we want to hide it behind a button or something.

jonasBoss:
> I very strongly prefer this variation:

I always felt the current UI is not very intuitive to navigate. This is why I rearranged some things:

The Idea is always to navigate the UI from top to bottom, and from left to right. So I put things like `apply` to the right, while `delete` is more on the left (relativly). 

I also tried to group the actions. `delete`, `copy` and `new` modify the preset, while `apply` and `autoload` are concerned with the injection. It is also a bit confusing that `Stop Injection` is not next to `apply` but I get the Idea why.

I split the editor in to 3 columns to provide a order in which things should be done:
 1. Select a mapping or the empty mapping
 2. Configure what to do: Key/Macro or Axis? and what is the Input?
 3. What is the output? What is the exact Macro, what are the settings for the axis.

This also gives some insights on that topic:
https://vanseodesign.com/web-design/direct-the-eye/

You are absolutely right with the spacing though.

I will work on making the UI functional first. I think we can have a better discussion on the Layout once it is actually usable and and most of the features are there.

sezanzeb:
I still don't understand 

image

Furthermore, I have to be able to understand what this is doing just by looking at it

jonasBoss:
@sezanzeb The Injection system is pretty much finished. The Unit test are passing (except for the macro I just changed) I only need to add more tests for all the new classes I implemented. 

I will work on the UI next. However, due to the changes I made to the preset and mapping that is more involved than I first anticipated. I have a few Ideas how I would go about implementing the changes. But I think it would be beneficial if we could go over the ideas and discuss them first. 

Do you have time for a zoom/discord/whatever call? I think that would be the easiest way.

sezanzeb:
> Do you have time for a zoom/discord/whatever call? I think that would be the easiest way.

Yes, next month

sezanzeb:
created beta branch

sezanzeb:
@clausbertels (https://github.com/sezanzeb/input-remapper/issues/7#issuecomment-751513986) might have some ideas on how to improve the user interface :). See https://github.com/sezanzeb/input-remapper/pull/263#issuecomment-1032057299 and below for screenshots of the current state

jonasBoss:
I just implemented the last back-end changes, this should now have the same functionality as the main branch plus some more. 
I think it is ready for merge into the beta branch. 
As we discussed, I will open new PRs for the work on the front-end.

sezanzeb:
what happened to tests like `test_hold_two`?

sezanzeb:
Can't find `test_filter_trigger_spam` in beta

sezanzeb:
Thanks for explaining!

as for `test_filter_trigger_spam`, looking back in history the test once looked like this

```py
        for i in range(1, 20):
            handle_keycode(_key_to_code, {}, InputEvent(*trigger, i), uinput)
```

so it was testing various values, like when slowly pressing a trigger. I don't know why it was changed to `1`.

sezanzeb:
Ah, because the old `should_map_event_as_btn` function was taking care of normalizing the value to 0 or 1 at some point.

sezanzeb:
> I still don't understand
> 
> image
> 
> at all. I don't even know where to start to ask questions what this is. sorry

@jonasBoss Thinking back, I sometimes feel bad because this was very rude, and I'm sorry. I was very grateful for your work, in case you didn't know, but bad at communicating (at least over text). If we had talked over discord or something I think it would have sounded a lot friendlier than it reads actually.

jonasBoss:
No worries, I never felt that you where rude. I made major changes to your work, so it was always clear to me that I had to explain my ideas and changes. I think that we worked well together and I am grateful that I could contribute to such a relevant project in a meaningful way.

PR #264, opened 2022-01-17
Use tempfiles instead of hardcoded /tmp

LunNova:
/tmp isn't guaranteed to be the correct place for temporary files in all environments, use tempfiles instead which checks appropriate environment variables.

Expected behavior with this change:

* Temp files end up in `$TMP/input-remapper-test<randomstring>`.
* * If $TMP isn't set, it falls back to the system default /tmp.
* Temp files are cleaned up after tests run if anything is left over
* Temp files won't end up in the current working directory

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -120,10 +120,6 @@ uinput_write_history_pipe = multiprocessing.Pipe()
 pending_events = {}
 
 
-if os.path.exists(tmp):
-    shutil.rmtree(tmp)
-
-
 def read_write_history_pipe():
     """convert the write history from the pipe to some easier to manage list"""
     history = []
--- a/tests/test.py
+++ b/tests/test.py
@@ -111,7 +111,7 @@ MIN_ABS = -(2 ** 15)
 MAX_ABS = 2 ** 15
 
 tmp_TemporaryDirectory = tempfile.TemporaryDirectory(prefix="input-remapper-test")
-tmp = str(tmp_TemporaryDirectory)
+tmp = tmp_TemporaryDirectory.name
 uinput_write_history = []
 # for tests that makes the injector create its processes
 uinput_write_history_pipe = multiprocessing.Pipe()
--- a/tests/test.py
+++ b/tests/test.py
@@ -110,8 +110,10 @@ START_READING_DELAY = 0.05
 MIN_ABS = -(2 ** 15)
 MAX_ABS = 2 ** 15
 
-tmp_TemporaryDirectory = tempfile.TemporaryDirectory(prefix="input-remapper-test")
-tmp = tmp_TemporaryDirectory.name
+# When it gets garbage collected it cleans up the temporary directory so it needs to
+# stay reachable while the tests are ran.
+temporary_directory = tempfile.TemporaryDirectory(prefix="input-remapper-test")
+tmp = temporary_directory.name
 uinput_write_history = []
 # for tests that makes the injector create its processes
 uinput_write_history_pipe = multiprocessing.Pipe()
--- a/tests/test.py
+++ b/tests/test.py
@@ -114,6 +114,7 @@ MAX_ABS = 2 ** 15
 # stay reachable while the tests are ran.
 temporary_directory = tempfile.TemporaryDirectory(prefix="input-remapper-test")
 tmp = temporary_directory.name
+
 uinput_write_history = []
 # for tests that makes the injector create its processes
 uinput_write_history_pipe = multiprocessing.Pipe()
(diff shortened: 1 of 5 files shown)


LunNova:
Could someone on a more usual linux distro than nixos check the tests still work okay with this?

jonasBoss:
on Manjaro KDE tests are passing, but
It creates a new directory called `<TemporaryDirectory '/` containing the directories  `tmp/input-remapper-test***'>/` inside the project folder. This is not cleaned up. It looks like the `''` are not interpreted correctly, which probably explains why the cleanup fails.

image

LunNova:
Okay now?

jonasBoss:
Yes.

tests/testcases/test_presets.py
```diff
@@ -172,7 +172,7 @@ def test_find_newest_preset_1(self):
         path = os.path.join(PRESETS, "Bar Device", "picture.png")
         os.mknod(path)
 
-        os.system("find /tmp/input-remapper-test/")
+        os.system("find " + tmp + "/")
```
sezanzeb, line 175: Formatstrings instead of string concatenation please.

```suggestion
        os.system(f"find {tmp}/")
```

tests/testcases/test_paths.py
```diff
@@ -36,15 +35,15 @@ def tearDown(self):
...
+        self.assertTrue(os.path.isfile(tmp + "/a/b/c/d/e"))
+        self.assertRaises(ValueError, lambda: touch(tmp + "/a/b/c/d/f/"))
 
     def test_mkdir(self):
-        mkdir("/tmp/b/c/d/e")
-        self.assertTrue(os.path.exists("/tmp/b/c/d/e"))
-        self.assertTrue(os.path.isdir("/tmp/b/c/d/e"))
+        mkdir(tmp + "/b/c/d/e")
+        self.assertTrue(os.path.exists(tmp + "/b/c/d/e"))
+        self.assertTrue(os.path.isdir(tmp + "/b/c/d/e"))
```
sezanzeb, line 46:
```suggestion
        self.assertTrue(os.path.isdir(os.path.join(tmp, "/b/c/d/e")))
```

here and everywhere else

tests/test.py
```diff
@@ -109,8 +110,8 @@ def join_children():
 MIN_ABS = -(2 ** 15)
 MAX_ABS = 2 ** 15
 
-
-tmp = "/tmp/input-remapper-test"
+tmp_TemporaryDirectory = tempfile.TemporaryDirectory(prefix="input-remapper-test")
+tmp = tmp_TemporaryDirectory.name
```
sezanzeb, line 114:
```suggestion
tmp = tempfile.TemporaryDirectory(prefix="input-remapper-test")
```

sezanzeb (changes requested):
Thanks for your contribution!

String concatenations are considered bad practice, formatstrings are much more readable.

To concatenate paths, `os.path.join` exists. https://stackoverflow.com/questions/13944387/why-use-os-path-join-over-string-concatenation

tests/test.py
LunNova: When it gets garbage collected it cleans up the temporary directory so it needs to stay reachable while the tests are ran.

sezanzeb:
```suggestion
# When it gets garbage collected it cleans up the temporary directory so it needs to
# stay reachable while the tests are ran.
temporary_directory = tempfile.TemporaryDirectory(prefix="input-remapper-test")
tmp = temporary_directory.name
```

tests/testcases/test_paths.py
sezanzeb: If you say you'd rather keep the string concatenation in those cases I'm fine with it

sezanzeb:
`with tempfile.TemporaryDirectory() as local_tmp:` this is nice. I'll make sure to remember this for future project

tests/testcases/test_presets.py
```diff
@@ -172,7 +172,7 @@ def test_find_newest_preset_1(self):
         path = os.path.join(PRESETS, "Bar Device", "picture.png")
         os.mknod(path)
 
-        os.system("find /tmp/input-remapper-test/")
+        os.system(f"find {tmp}/")
```
LunNova, line 175: Oh, this will probably not work correctly if tmp has spaces in it now. Looks like it's just for debug logging but I should still fix that.

LunNova: Would changing this to `subprocess.call(["find", tmp])` be fine?

sezanzeb: oh, this is leftover from debugging. You can remove this line

sezanzeb:
Thanks!

PR #265, opened 2022-01-17
Use non-zero exit code if tests fail

LunNova:
Lets tools running the tests know if any failed without having to parse the output.

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -35,7 +35,6 @@ if __name__ == "__main__":
     import tests.test
 
     tests.test.main()
-    sys.exit(0)
 
 import shutil
 import time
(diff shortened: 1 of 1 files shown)


tests/test.py
```diff
@@ -35,7 +35,6 @@
     import tests.test
 
     tests.test.main()
-    sys.exit(0)
```
sezanzeb, line 38: @jonasBoss was using the sys.exit here to prevent importing this file twice or something

jonasBoss: since we have the sys.exit at the end of main() this is just fine. we don't need it here, I just added it because it was missing before. Reporting if the test passed makes totally sense.

sezanzeb:
Thank you

Issue #266, opened 2022-01-17
Run unittests via ci pipeline


sezanzeb:
possibly related to https://github.com/sezanzeb/input-remapper/issues/246

LunNova:
Happy to help set up github actions to do this once it's split.

sezanzeb:
Done

PR #267, opened 2022-01-18
Add github actions workflows

LunNova:
WIP for #266 

What works: linting, also have it [adding PR comments.](https://github.com/LunNova/input-remapper/pull/2#discussion_r786410522)

Doesn't work: running tests. Some non-test code tries to chown things either in /tmp or /proc and fails.

```
   File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/inputremapper/gui/reader.py", line 74, in connect
    self._results = Pipe(f"/tmp/input-remapper-{USER}/results")
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/inputremapper/ipc/pipe.py", line 72, in __init__
    chown(f"{fds_dir}{self._fds[0]}")
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/inputremapper/paths.py", line 35, in chown
    shutil.chown(path, user=USER, group=USER)
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/shutil.py", line 1071, in chown
    os.chown(path, _user, _group)
PermissionError: [Errno 1] Operation not permitted: '/proc/3695/fd/10'
```

```
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/inputremapper/gui/reader.py", line 74, in connect
    self._results = Pipe(f"/tmp/input-remapper-{USER}/results")
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/inputremapper/ipc/pipe.py", line 59, in __init__
    mkdir(os.path.dirname(path))
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/inputremapper/paths.py", line 75, in mkdir
    chown(path)
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/inputremapper/paths.py", line 35, in chown
    shutil.chown(path, user=USER, group=USER)
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/shutil.py", line 1071, in chown
    os.chown(path, _user, _group)
PermissionError: [Errno 1] Operation not permitted: '/tmp/input-remapper-runner'
```

Will look into it later, might have to wait for the weekend.

### DIFF
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -43,4 +43,4 @@ jobs:
           REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           reviewdog -list
-          reviewdog -tee -runners=mypy,pylint,black -reporter=${{ env.REWIEVDOG_REPORTER }} -fail-on-error=false
+          reviewdog -tee -runners=mypy,black -reporter=${{ env.REWIEVDOG_REPORTER }} -fail-on-error=false
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -1,0 +1,46 @@
+---
+name: reviewdog
+# run reviewdog for PR only because "github-check" option is failing :(
+# https://github.com/reviewdog/reviewdog/issues/924
+on: [pull_request]
+
+jobs:
+  reviewdog_python:
+    name: reviewdog - Python lint
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.10"]
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - uses: reviewdog/action-setup@master
+        with:
+          reviewdog_version: latest
+      - name: Install dependencies
+        shell: bash
+        run: |
+          scripts/ci-install-deps.sh
+          pip install flake8 pylint mypy black
+      - name: Set env for PR
+        if: github.event_name == 'pull_request'
+        shell: bash
+        run: echo "REWIEVDOG_REPORTER=github-pr-review" >> $GITHUB_ENV
+
+      - name: Set env for push
+        if: github.event_name != 'pull_request'
+        shell: bash
+        run: echo "REWIEVDOG_REPORTER=github-check" >> $GITHUB_ENV
+
+      - name: Run reviewdog
+        shell: bash
+        env:
+          REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          reviewdog -list
+          reviewdog -tee -runners=mypy,pylint,black -reporter=${{ env.REWIEVDOG_REPORTER }} -fail-on-error=false
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,28 @@
+name: Test
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    continue-on-error: true
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+    strategy:
+      matrix:
+        python-version: ["3.7", "3.10"] # min and max supported versions?
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        # Install deps as root since we will run tests as root
+        sudo scripts/ci-install-deps.sh
+        sudo pip install .
+    - name: Run tests
+      run: |
+        # FIXME: Had some permissions issues, currently worked around by running tests as root
+        mkdir test_tmp
+        TMPDIR="$(realpath test_tmp)" sudo python tests/test.py --start-dir unit
--- a/scripts/ci-install-deps.sh
+++ b/scripts/ci-install-deps.sh
@@ -3,16 +3,12 @@
 set -xeuo pipefail
 
 # native deps
-# gettext required to generate translations
-sudo apt-get install -y gettext
+# gettext required to generate translations, others are python deps
+sudo apt-get install -y gettext python3-evdev python3-pydbus python3-pydantic
 
 # ensure pip and setuptools/wheel up to date so can install all pip modules
 python -m pip install --upgrade pip
 pip install wheel setuptools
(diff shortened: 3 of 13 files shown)


sezanzeb:
There is a bunch of stuff that gets reported by pylint but that I chose to ignore for reasons. It even reports one error which rather seems like a bug in pylint. So does pylint work for this project?

LunNova:
https://github.com/LunNova/input-remapper/actions/runs/1767220593
https://github.com/LunNova/input-remapper/actions/runs/1767220595

Example test runs.
Some unit tests failing on CI:

```
======================================================================
ERROR: test_data_usr (unit.test_data.TestData)
----------------------------------------------------------------------
  File "/home/runner/work/input-remapper/input-remapper/tests/unit/test_data.py", line 40, in test_data_usr
    self.assertTrue(get_data_path().startswith("/usr/"))
  File "/usr/local/lib/python3.8/dist-packages/inputremapper/data.py", line 82, in get_data_path
    sys.exit(1)
SystemExit: 1

======================================================================
FAIL: test_numlock (unit.test_injector.TestInjector)
----------------------------------------------------------------------
  File "/usr/lib/python3.8/unittest/async_case.py", line 65, in _callTestMethod
    self._callMaybeAsync(method)
  File "/usr/lib/python3.8/unittest/async_case.py", line 84, in _callMaybeAsync
    ret = func(*args, **kwargs)
  File "/home/runner/work/input-remapper/input-remapper/tests/unit/test_injector.py", line 247, in test_numlock
    self.assertEqual(not before, is_numlock_on())
AssertionError: True != None
```

Black is also failing but that's right because there are some files with formatting issues currently.

sezanzeb:
Thanks! I moved things a bit around on main now, so that the two tests above won't cause issues for the ci anymore.

LunNova:
Another example PR: https://github.com/LunNova/input-remapper/pull/3

Everything's passing now except black, and reviewdog added a comment for something pylint found.

The on-push pylint run is currently set to always pass, so it's either useless or for information only depending on if you expect people to open up and look at its test output.

https://github.com/LunNova/input-remapper/runs/5039604773?check_suite_focus=true

It's set that way because there are quite a few warnings it's finding on main already so it would always fail. I may have set something up wrong.

LunNova:
I also added the option to run `python tests/test.py --start-dir unit` to make it easy to run all the unit tests without keeping a list.

.github/workflows/reviewdog.yml
```diff
@@ -0,0 +1,46 @@
...
+        shell: bash
+        run: echo "REWIEVDOG_REPORTER=github-check" >> $GITHUB_ENV
+
+      - name: Run reviewdog
+        shell: bash
+        env:
+          REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          reviewdog -list
+          reviewdog -tee -runners=mypy,pylint,black -reporter=${{ env.REWIEVDOG_REPORTER }} -fail-on-error=false
```
LunNova, line 46: If any of these linters end up not being useful, remove from the list here to disable.

.github/workflows/test.yml
```diff
@@ -0,0 +1,28 @@
...
+on: [push, pull_request]
+
+jobs:
+  build:
+    continue-on-error: true
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+    strategy:
+      matrix:
+        python-version: ["3.7", "3.10"] # min and max supported versions?
```
LunNova, line 12: Unit tests run on min and max python versions you want to support

scripts/ci-install-deps.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+# Called from multiple CI pipelines in .github/workflows
+set -xeuo pipefail
+
+# native deps
+# FIXME: this is a lot and most aren't actually used in testing but tests will crash without due to GtkSource require
```
LunNova, line 6: This is slow but it does work.

With some further changes to avoid importing Gtk things during unit test runs this list could be a lot smaller, I'm not sure if that's worth doing.

LunNova:
Looks like those black changes got accidentally reverted in https://github.com/sezanzeb/input-remapper/commit/3732d9e1e65dc1c3c97b26c56f5e92cdb8e6a95f so that's still failing

:sweat_smile:

scripts/ci-install-deps.sh
sezanzeb: pushed a change to main, gtk dependencies are not required for unit tests anymore

.github/workflows/reviewdog.yml
sezanzeb: Lets remove pylint, it is a nice tool to learn good practices, but the list of stuff I'm ignoring on purpose is too long at this point to be really useful.

.github/workflows/test.yml
sezanzeb: Look good, I hope it works on python 3.7. But I think I tried to only use features of 3.6 actually, but I lost track of it at this point.

sezanzeb:
> Looks like those black changes got accidentally reverted in https://github.com/sezanzeb/input-remapper/commit/3732d9e1e65dc1c3c97b26c56f5e92cdb8e6a95f so that's still failing

Can you run the required black version on the code and include its changes in this PR?

scripts/ci-install-deps.sh
LunNova: Nice, it's a lot faster now. I had to remove one more Gtk dependency in test_daemon.py which was unused.

.github/workflows/reviewdog.yml
LunNova: Done

LunNova:
Ran black as well

sezanzeb:
Thanks! Is there anything missing?

At which point will we see ci results in the commit history?

LunNova:
You'll see CI results on the checkmark next to a commit: 
image

For PRs they will also show up at the bottom of the MR like this:

image

CI will only run when a branch or PR is pushed/updated. It won't go back and run against older commits. If CI fails after you push, github will email you. If that isn't helpful you can turn it off here https://github.com/settings/notifications#github-actions-heading

sezanzeb:
Ok. Shouldn't there also be ci runs visible there:

image

?

LunNova:
They won't show up on this PR because there are no actions in the `main` branch. I think this is normal behavior for a PR adding actions.

LunNova:
Yeah, let's go :rocket: 

If one of the workflows is causing problems you can disable it through the github UI without needing to make a commit at https://github.com/sezanzeb/input-remapper/actions/workflows/lint.yml with the menu near the right.

sezanzeb:
Thanks a lot!

PR #275, opened 2022-01-29
Renaming and minor stuff

jonasBoss:
This contains part of the changes from #263 pretty much everything that is not related to the new injection system:
 - renamed Mapping to Preset
 - renamed the global `config` object to `global_config`
 - moved config files to `configs` directory
 - renamed the `custom_mapping` (GUI) to `active_preset`
 - replaced `Key` with `InputEvent` and `EventCombination`

I think it is a good Idea to merge this first, so that we can eliminate a lot of errors which might be caused by the renaming of stuff.
It also should help with code Review on #263 since it reduces the size dramaticly


sezanzeb (commented):
I'm working on fixing tests on my computer by optimizing a few things before I'll merge this. Just to make sure tests for the changes are also passing on my machine.

sezanzeb:
tests are working again on main for me, but there are some conflicts now

jonasBoss:
I resolved them. Everything should work again.

sezanzeb:
thanks!

Issue #278, opened 2022-02-02
Checked out repo, app will not launch due to keyword argument error


lbattraw:
I'm running Ubuntu 21.10, checked out repo today, when launching the app or starting the service it fails with the error:

# input-remapper-gtk
input-remapper-gtk 1.4.0 b3e1e4ca1976bb420017c737525151ac37099e6e https://github.com/sezanzeb/input-remapper
python-evdev 1.4.0
Traceback (most recent call last):
  File "/bin/input-remapper-gtk", line 70, in <module>
    from inputremapper.gui.user_interface import UserInterface
  File "/usr/local/lib/python3.9/dist-packages/inputremapper/gui/user_interface.py", line 32, in <module>
    from inputremapper.input_event import InputEvent
  File "/usr/local/lib/python3.9/dist-packages/inputremapper/input_event.py", line 30, in <module>
    @dataclass(frozen=True, slots=True)
TypeError: dataclass() got an unexpected keyword argument 'slots'

I don't know much about this type of error, however removing that argument allows it to launch, line 30 in  /usr/local/lib/python3.9/dist-packages/inputremapper/input_event.py works as:

@dataclass(frozen=True)

MostHated:
I keep getting this as well. I had key-mapper installed prior, not sure if it is related. I removed it first. I tried simply installing the premade deb as well, but it didn't work either.

sezanzeb:
upgrading python should do. Other than that

@jonasBoss the slots argument is only supported since python3.10

sezanzeb:
Why does the deb not work?

jonasBoss:
oops, I was not aware of that. I guess we should do testing in a 3.8 venv. Maybe also in the upcoming CI @LunNova 
We can just remove the slots argument for now.

But it might be a good idea to leave a note to revert this change once 3.10 arrives in common distros. With #263 we end up creating several thousand `InputEvent` objects a second. Depending on the polling rate of input devices.

sezanzeb:
The original issue should be fixed now.

@MostHated If the deb does not work, please open another ticket and explain the problem in more detail. Since the change that added the DataClass is not in the deb file it is a different issue.

LunNova:
> Maybe also in the upcoming CI @LunNova

Currently it's set to test on 3.7 and 3.10: https://github.com/sezanzeb/input-remapper/pull/267/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R12

PR #279, opened 2022-02-02
fix python 3.8 compatability #278

jonasBoss:


jonasBoss:
that did not work, for some reason it there are 55 files changed

sezanzeb:
probably because I'm squashing all changes (including mine)

jonasBoss:
makes sense

sezanzeb:
does it? Why would it show

image

as a change

jonasBoss:
I guess that is the last common commit in the history. Git does not know you merged all my commits via the PR

PR #288, opened 2022-02-07
Improve i18n

jose1711:


sezanzeb:
```
  File "/mnt/data/Code/input-remapper/inputremapper/gui/user_interface.py", line 116, in wrapped
    return func(self, *args, **kwargs)
  File "/mnt/data/Code/input-remapper/inputremapper/gui/user_interface.py", line 431, in on_restore_defaults_clicked
    self.show_status(CTX_APPLY, _("Applied the system default"))
TypeError: 'tuple' object is not callable
```

jose1711:
> on_restore_defaults_clicked

thank you, corrected

sezanzeb:
There are more issues like this.

I'd recommend running the test to find them all `python3 tests/test.py integration.test_gui`

jose1711:
> There are more issues like this.
> 
> I'd recommend running the tests to find them all `python3 tests/test.py integration.test_gui`

Thank you, tests seem to be happy now.

sezanzeb (changes requested):

inputremapper/gui/editor/editor.py
```diff
@@ -23,9 +23,22 @@
...
 from gi.repository import Gtk, GLib, Gdk
 
+APP_NAME = 'input-remapper'
+LOCALE_DIR = os.path.join(get_data_path(), 'lang')
+
+locale.bindtextdomain(APP_NAME, LOCALE_DIR)
+locale.textdomain(APP_NAME)
+
+translate = gettext.translation(APP_NAME, LOCALE_DIR, fallback=True)
+_ = translate.gettext
```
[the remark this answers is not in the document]
jose1711: Duplicate code moved to `gettext.py` as suggested.

sezanzeb:
You don't need to force-push by the way, I'm going to squash all commits anyway

inputremapper/gui/gettext.py
```diff
@@ -0,0 +1,34 @@
...
+import os.path
+import gettext
+import locale
+from inputremapper.configs.data import get_data_path
+from argparse import ArgumentParser
+
+APP_NAME = "input-remapper"
+LOCALE_DIR = os.path.join(get_data_path(), "lang")
+
+locale.bindtextdomain(APP_NAME, LOCALE_DIR)
```
sezanzeb, line 30: @LunNova the linter fails here, but `locale.bindtextdomain` exists. What can we do about it?

bin/input-remapper-gtk
```diff
@@ -24,27 +24,15 @@
 
 import sys
 import atexit
-import gettext
-import locale
-from inputremapper.configs.data import get_data_path
-import os.path
 from argparse import ArgumentParser
+from inputremapper.gui.gettext import _, LOCALE_DIR
```
sezanzeb, line 28:
```suggestion
from inputremapper.gui.gettext import _
```

sezanzeb (changes requested):

bin/input-remapper-gtk
jose1711: Please note that `LOCALE_DIR` is used in the debug message.

jose1711:
> You don't need to force-push by the way, I'm going to squash all commits anyway. Each time you do a force push I need to make a hard reset locally

Thank you, noted down. I was taught to do force-push in the past for other projects.

bin/input-remapper-gtk
sezanzeb: ah, whoops

sezanzeb:
why is it closed?

inputremapper/gui/gettext.py
sezanzeb: image

but it seems this is optional, the pipeline passes: 
image

sezanzeb: ah, this is reviewdog and it seems to be something different from the pipelines I have seen in the past

jose1711:
> why is it closed?

:question: :shrug:

inputremapper/gui/gettext.py
LunNova: I raised an issue, that should not have been marked as an error, some definitions are missing. https://github.com/python/typeshed/issues/7164

Looks like those functions aren't often supposed to be used, the docs say to use `gettext.bindtextdomain` instead in most cases. Would that work here or is working with Gtk one of the cases where we need to use these? I'm not sure. https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Doc/library/locale.rst#access-to-message-catalogs

sezanzeb: ok, thank you!

no clue about t18n, maybe @jose1711 can answer this question

PR #289, opened 2022-02-07
Update Slovak translation.

jose1711:


po/sk_SK.po
```diff
@@ -18,187 +18,243 @@ msgstr ""
...
-#: data/input-remapper.glade:898
+#: data/input-remapper.glade:1070
 msgid "About"
 msgstr "O programe"
 
-#: data/input-remapper.glade:209
+#: inputremapper/gui/user_interface.py:609
+#, python-format
+msgid "Applied preset %s"
+msgstr "Použité prednastavenie %1"
```
sezanzeb, line 32: `%1`, shouldn't this be `%s`?

jose1711: correct. force-pushed together with a few other issues.

sezanzeb:
Thank you

PR #331, opened 2022-03-07
Debounce when typing

kearkyle:
Moved debounce to utils.py, and added debounce to check_macro_syntax when TextBuffer is in "changed" condition.

### DIFF
--- a/inputremapper/gui/user_interface.py
+++ b/inputremapper/gui/user_interface.py
@@ -206,7 +206,7 @@ class UserInterface:
         source_view = self.get("code_editor")
 
         # throw debounce on typing
-        source_view.get_buffer().connect("changed", self.check_macro_syntax)
+        source_view.get_buffer().connect("changed", self.check_on_typing)
 
         # if any of the next steps take a bit to complete, have the window
         # already visible (without content) to make it look more responsive.
@@ -473,6 +473,9 @@ class UserInterface:
             status_bar.set_tooltip_text(tooltip)
 
     @debounce
+    def check_on_typing(self, *_):
+        self.check_macro_syntax()
+
     def check_macro_syntax(self):
         """Check if the programmed macros are allright."""
         self.show_status(CTX_MAPPING, None)
--- a/inputremapper/gui/user_interface.py
+++ b/inputremapper/gui/user_interface.py
@@ -206,7 +206,7 @@ class UserInterface:
         source_view = self.get("code_editor")
 
         # throw debounce on typing
-        source_view.get_buffer().connect("changed", self.check_on_typing)
+        source_view.get_buffer().connect("changed", self.check_macro_syntax)
 
         # if any of the next steps take a bit to complete, have the window
         # already visible (without content) to make it look more responsive.
@@ -473,13 +473,12 @@ class UserInterface:
             status_bar.set_tooltip_text(tooltip)
 
     @debounce
-    def check_on_typing(self, *_):
-        self.check_macro_syntax()
-
     def check_macro_syntax(self):
         """Check if the programmed macros are allright."""
         self.show_status(CTX_MAPPING, None)
+        print("check_macro_syntax")
         for key, output in active_preset:
+            print("checking", key, output)
             output = output[0]
             if not is_this_a_macro(output):
                 continue
@@ -787,7 +786,7 @@ class UserInterface:
 
             # checking macros is probably a bit more expensive, do that if
             # the regular mappings are allright
-            self.check_macro_syntax()
+            # self.check_macro_syntax()
 
     def on_about_clicked(self, button):
         """Show the about/help dialog."""
--- a/inputremapper/gui/user_interface.py
+++ b/inputremapper/gui/user_interface.py
@@ -147,7 +147,6 @@ class UserInterface:
 
         self.group = None
         self.preset_name = None
-        self.debounce_timeout = 100
 
         global_uinputs.prepare()
         css_provider = Gtk.CssProvider()
@@ -167,6 +166,7 @@ class UserInterface:
         self.builder = builder
 
         self.editor = Editor(self)
+        self.debounce_timeout = 100
 
         # set up the device selection
         # https://python-gtk-3-tutorial.readthedocs.io/en/latest/treeview.html#the-view
@@ -204,8 +204,9 @@ class UserInterface:
 
         # attaching source_view to editing window
         source_view = self.get("code_editor")
+
         # throw debounce on typing
-        source_view.get_buffer().connect("changed", self.check_on_typing)
+        source_view.get_buffer().connect("changed", self.check_macro_syntax)
 
         # if any of the next steps take a bit to complete, have the window
         # already visible (without content) to make it look more responsive.
@@ -472,10 +473,6 @@ class UserInterface:
             status_bar.set_tooltip_text(tooltip)
 
     @debounce
-    def check_on_typing(self):
-        """Check if the programmed macros are allright."""
-        self.check_macro_syntax
-
     def check_macro_syntax(self):
         """Check if the programmed macros are allright."""
         self.show_status(CTX_MAPPING, None)
--- a/inputremapper/gui/user_interface.py
+++ b/inputremapper/gui/user_interface.py
@@ -206,7 +206,7 @@ class UserInterface:
         source_view = self.get("code_editor")
 
         # throw debounce on typing
-        source_view.get_buffer().connect("changed", self.check_macro_syntax)
+        source_view.get_buffer().connect("changed", self.check_on_typing)
 
         # if any of the next steps take a bit to complete, have the window
         # already visible (without content) to make it look more responsive.
@@ -473,12 +473,14 @@ class UserInterface:
             status_bar.set_tooltip_text(tooltip)
 
     @debounce
+    def check_on_typing(self, *_):
+        self.check_macro_syntax()
+        self.editor.gather_changes_and_save()
+
     def check_macro_syntax(self):
         """Check if the programmed macros are allright."""
         self.show_status(CTX_MAPPING, None)
-        print("check_macro_syntax")
         for key, output in active_preset:
-            print("checking", key, output)
             output = output[0]
             if not is_this_a_macro(output):
                 continue
--- a/inputremapper/gui/user_interface.py
+++ b/inputremapper/gui/user_interface.py
@@ -72,6 +72,7 @@ from inputremapper.gui.utils import (
     CTX_APPLY,
     CTX_WARNING,
     gtk_iteration,
+    debounce,
 )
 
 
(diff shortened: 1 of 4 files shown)


sezanzeb:
Have you tried it on your computer, is it working?

sezanzeb:
you probably didn't install your own changes correctly, because this is the old behaviour

kearkyle:
I notice it is the same old behavior, but I uninstalled and installed again couple times it would still perform like that, if I commented all check_macro_syntax out then instructions don't show, but if I keep my changes and commented the old behavior's syntax check, it would still behave like it did before

sezanzeb:
carefully look through your changes once more, try to find your mistake. Your code is not doing anything as of now

kearkyle:
I have a question about this part `source_view = self.get("code_editor")`. Is code_editor the user text input or is it not? I believe it should be but since it is not doing anything from my changes so I wonder if that could be the issue.

sezanzeb:
it is the code editor, yes.

You can easily figure out if your signal is firing by doing something like

```py
    def check_macro_syntax(self):
        """Check if the programmed macros are allright."""
        print("check_macro_syntax")
        ...
```

then running it and observing the output.

And then maybe put some more logs into it to figure out what is going wrong. like

```py
    def check_macro_syntax(self):
        """Check if the programmed macros are allright."""
        print("check_macro_syntax")
        self.show_status(CTX_MAPPING, None)
        for key, output in active_preset:
            print("checking", key, output)
```

do you notice anything odd in the output?

kearkyle:
My bad I just came back from long vacation. So I added the print signals and this is what I got:

https://user-images.githubusercontent.com/43527143/160527250-eed07ede-6e32-445a-a2ba-0822384fc476.mp4


https://user-images.githubusercontent.com/43527143/160527262-bd2cf136-2f46-4b3c-a8f5-9d485607a822.mp4

It does look like it is not printing out the most recent keyboard events, but the last one when clicking off and on the application again. I did make sure to comment out the `self.check_macro_syntax()` in save_preset so I don't think that is triggering the function by any chance. Should I reduce debounce_timeout ? It is currently set as 100 equal to autocompletion.py

sezanzeb:
if `for key, output in active_preset` gives you old text no matter how much you type, then the `__iter__` function of the `Preset` class must be having old information.

You'll have to figure out why this information is old, and why it is up-to-date when leaving and entering the text area

kearkyle:
So I added `self.editor.gather_changes_and_save()` and now active_preset is getting the input from user. When printing out it would not print the most current but the last, so If I enter **ab**, it would print **a**, **abc** would print **ab** and so on but syntax checking is still current if we notice as a type if_single() and it display that exact same thing under the left corner for syntax checking. I also noticed that `check_macro_syntax()` is working when pressing enter at end of line instead of after the user end typing. Is that supposed to be the normal behavior of `check_macro_syntax()`? 

https://user-images.githubusercontent.com/43527143/161000938-1245f017-9a62-482a-a66f-d8f06361162a.mp4



https://user-images.githubusercontent.com/43527143/161000985-ee12df05-7991-4134-aae9-13a098e1d6b2.mp4

sezanzeb:
right now you are checking the syntax, and then you are calling `gather_changes_and_save`. This needs to be the other way around

inputremapper/gui/user_interface.py
```diff
@@ -465,6 +472,11 @@ def show_status(self, context_id, message, tooltip=None):
             status_bar.push(context_id, message)
             status_bar.set_tooltip_text(tooltip)
 
+    @debounce
+    def check_on_typing(self, *_):
```
sezanzeb, line 476: Try to add a docstring. It should have the format of

```py
    def check_on_typing(self, *_):
        """Do this and that."""
```

The line with the comment should not be longer than 88 characters. A short sentence that fits in a single line.

Docstrings are super important. They will save you once you have to look at old code and don't know anymore what it is doing. Having to write a very short docstring for each function makes you think twice if that function is really doing something that can easily be explained. If it is hard to explain what a function is doing, it might need some redesigning.

sezanzeb:
Good.

To fix the problem in the pipeline:

image

you'll have to install black (https://github.com/psf/black) and run `black .`. This will reformat your files. Afterwards commit the changes.

kearkyle:
I did that just now. Can you run the check again to see if it passes now?

kearkyle:
Just making sure, so I make another function that is similar to `test_check_macro_syntax` but replacing `self.user_interface.save_preset()` with `self.editor.set_symbol_input_text(symbol)` ?

sezanzeb:
You basically have to write a test that simulates typing, which should then trigger a syntax check.

the `test_check_macro_syntax` might be a starting point. But I realized that if you just copy it and change those two lines you won't test debouncing. So maybe just use that test as an example on how to write tests.

Here is an idea on how the test might look like:

1. insert broken macro
2. expect no error message to be shown, because the debouncing will wait for more input
3. wait for 110ms
4. expect an error message to be shown now

we need tests to make sure features don't suddenly break. Every software project needs automatic tests. Those without will definitely run into problems at some point. If I was to write software without tests at work I would have no guarantee that our product still works, and every deployment would be a huge gamble.

kearkyle:
I'm still working on the debounce test. But so far when I run test.py this are all the errors and failures that I get:

sezanzeb:
It is better if you copy the text from the console and paste it here instead of making screenshots. It should include the command that you used to run tests.

Those tests are not failing for me. `unit.test_macros.TestMacros.test_2` might be failing because your device is not fast enough, don't worry about that one.

It appears your current working directory is not correct. You should be in the root of the github project when running tests.

kearkyle:
I have a question about the test for debounce. So the test would pass for me if the code is as below:
image
But if it should gives nothing at start, shouldn't `self.assertTrue(error_icon.get_visible())` stay on the bottom after `time.sleep(0.11)` and `self.assertFalse(error_icon.get_visible())`. Also for this line 
`self.assertIsNone(tooltip)` for some reason after 110ms it is returning None on the tooltip for me. Is my implementation of the test wrong ? Do I need to `save_preset` since it is already done when user type in the text field?

sezanzeb:
please don't post screenshots, copy and paste the code instead and wrap it in ` ``` `

like this:

```py
def foo(bar):
  return "qux"
```

I have a hard time understanding what "stay on the bottom" and "stay on top" mean

sezanzeb:
I can't tell what the issues is. Some tipps that might help understanding the problem:
- what are the tooltip text and icon visibility before you call `set_symbol_input_text`?
- what is in the tooltip text after `set_symbol_input_text`? Either use your IDEs debugger or add print statements to better understand what is going on.
- add a `gtk_iteration` after `set_symbol_input_text`, does that change anything?

kearkyle:
I mean supposing we `set_symbol_input_text(foo)`. The debounce is supposed to work this way: 
```
tooltip = status.get_tooltip_text().lower()
self.assertNotIn("foo", tooltip)     #tooltip = "Service has not started"
self.assertFalse(error_icon.get_visible())        #this should be False at start?
self.assertFalse(warning_icon.get_visible())   #false
``` 
and after 110ms it should be something like this:
```
tooltip = status.get_tooltip_text()
self.assertIn("foo", tooltip)                              
self.assertTrue(error_icon.get_visible())         #error icon visible after debounce
self.assertFalse(warning_icon.get_visible())
```
Is that supposed to be the correct behaviour of debounce ? Currently having those would fail the test, and it would pass for the following:
```
def test_debounce_check_on_typing(self):
        status = self.user_interface.get("status_bar")
        error_icon = self.user_interface.get("error_status_icon")
        warning_icon = self.user_interface.get("warning_status_icon")

        self.editor.set_symbol_input_text("foo")
        gtk_iteration()
        tooltip = status.get_tooltip_text().lower()
        self.assertNotIn("foo", tooltip)
        self.assertTrue(error_icon.get_visible())
        self.assertFalse(warning_icon.get_visible())

        time.sleep(0.11)
        gtk_iteration()
        
        tooltip = status.get_tooltip_text()
        self.assertIsNone(tooltip)         #tooltip is None here ?
        self.assertFalse(error_icon.get_visible())
        self.assertFalse(warning_icon.get_visible())

        self.assertEqual(self.editor.get_symbol_input_text(), "foo")
```
Notice the difference of error_icon before and after `time.sleep(0.11)`

sezanzeb:
after waiting for 110 milliseconds it should have triggered the function that you added, and therefore should have added tooltip text as far as I know.

You'll need to spend some time debugging this. I can tell why this isn't working from here in this case

sezanzeb:
I just ran tests and `test_check_macro_syntax` only fails on your branch, it doesn't fail on main.

You can run single tests via `python3 tests/test.py integration.test_gui.TestGui.test_check_macro_syntax`

sezanzeb:
a good tactic to figure out why existing tests are failing is to revert your changes step by step until they work again. Then you know which line causes the breakage.

kearkyle:
So I cloned and applied the changes to version 1.4.2 and everything looks good except for test_numlock() which gives me the following:
======================================================================
FAIL: test_numlock (tests.integration.test_numlock.TestNumlock)
----------------------------------------------------------------------
  File "/home/vchen7/Github/main/input-remapper/tests/integration/test_numlock.py", line 41, in test_numlock
    self.assertEqual(not before, is_numlock_on())
AssertionError: False != True

----------------------------------------------------------------------
Ran 307 tests in 78.593s
I didn't copy paste the whole test result because it's too large but I'm on the root directory for sure. 
Also for the debounce test implementation so far I came up with this:
```
def test_debounce_check_on_typing(self):
        status = self.user_interface.get("status_bar")
        error_icon = self.user_interface.get("error_status_icon")
        warning_icon = self.user_interface.get("warning_status_icon")

        print("Second check")
        self.add_mapping_via_ui(EventCombination([EV_KEY, 10, 1]), "if()")
        tooltip = status.get_tooltip_text()
        gtk_iteration()
        self.assertIn("if", tooltip)
        self.assertTrue(error_icon.get_visible())
        self.assertFalse(warning_icon.get_visible())
        
        time.sleep(0.11)
        gtk_iteration()
        
        print("third check")
        tooltip = status.get_tooltip_text()
        self.assertIn("if", tooltip)       
        self.assertTrue(error_icon.get_visible())
        self.assertFalse(warning_icon.get_visible())

        self.assertEqual(self.editor.get_symbol_input_text(), "if()")
```
I notice error_icon is visible and `print(tooltip)` will show **if** right after **if** is inserted on the text field. I'm still looking into this because we want them to show that after 110ms, not right after **if** is inserted into the text field.

sezanzeb:
I quickly completed this feature and pushed it to main.

Thanks for your contribution! :)

sezanzeb:
I screwed it up a little bit, you are not listed in [githubs contributors list](https://github.com/sezanzeb/input-remapper/graphs/contributors list) now :(

I'm really sorry for that, I hope this isn't important to you

kearkyle:
Hi @sezanzeb no worries at all, I couldn't do much and wasn't unable to do the test for that. I'm glad you solved it.

Issue #336, opened 2022-03-12
Error in service when injecting


basta:
This error occurs in `systemctl status input-remapper` when trying to inject a preset. 
```
Traceback (most recent call last):
   File "/usr/lib64/python3.10/multiprocessing/process.py", line 315, in _bootstrap
     self.run()
   File "/usr/local/lib/python3.10/site-packages/inputremapper/injection/injector.py", line 341, in run
     input_props=source.input_props(),
 AttributeError: 'InputDevice' object has no attribute 'input_props'

```
The app itself hangs on `Starting injection...`
Using Fedora 35, python 3.10, input-remapper version 1.4.1.

Not sure what additional debug info is needed, will update if necessary

basta:
As a shot in the dark I just tried removing the offending argument `input_props=source.input_props()` and injections started to work.

sezanzeb:
Thanks for reporting, I'll add some extra checks to avoid the issue

sezanzeb:
There was one more problem with this, fix is on main

sezanzeb:
What is you python-evdev version? See `input-remapper-control --version`

basta:
It's 1.1.2. Full log:
```
input-remapper 1.4.1 0cac61149c298541a3860521436954ee8747027f https://github.com/sezanzeb/input-remapper
python-evdev 1.1.2
```

sezanzeb:
yeah, please upgrade your python-evdev version instead

basta:
Sorry about that. 
But it should be the latest version in fedora's package manager.
Maybe a warning in the installation instructions is in order? Not sure if this issue is on my end

sezanzeb:
something about it is in the making: https://github.com/sezanzeb/input-remapper/pull/346

fedoras package manager seems to be very outdated in this regards.

sezanzeb:
Try `sudo pip install evdev`

basta:
That didn't help. The requirement was already satisfied by version 1.1.0.

But pip `install evdev==1.4.0` (last version) did. I don't understand pip packages enough to debug this. Maybe it didn't update the version because of some dependency from other package?

basta:
Upon deleting and reinstalling evdev using `pip install evdev` the latest version was installed correctly, so I think the issue was on my end. As a feature suggestion maybe some warnings in the console would help prevent this?

You've made an awesome software, thanks for your time.

PR #341, opened 2022-03-17
Fix for missing input_props attribute

sezanzeb:
Fixes https://github.com/sezanzeb/input-remapper/issues/336


sezanzeb:
Reverted in https://github.com/sezanzeb/input-remapper/pull/346

Issue #83, opened 2021-04-09
rpm-package


Blaimi:
I'd love to see this project as an rpm-package for fedora.

Blaimi:
I startet at https://github.com/Blaimi/key-mapper/tree/83_rpm-package :)

sezanzeb:
Cool, but bear in mind that I can't test it because I don't use any fedora based distro

When would I need to worry about potentially breaking the rpm build with changes to key-mapper?

Blaimi:
As long as you don't depend on newer versions of the python-libraries as they are in the repositories the should not be a problem. Since fedora has a faster release cycle than ubuntu (fedora has a ~6 month release cycle like ubuntu but without LTS-steps) I don't assume problems if the project fits the latest ubuntu LTS. Once the package is in the official repositories you cannot upload newer versions anyway for a certain release (i.e. if ubuntu 21.10 comes with key-mapper 1.0.0, you can upgrade the package to 1.1 not before ubuntu 22.04), but you can publish it in backports or a ppa.

I think if the fedora-package has problems with dependencies, you will have them on ubuntu as well. These problems should be solved when they pop up. That's another reason why I mentioned to split the packaging- from the source-codein #40. This way the package-stuff could even be done on the gitlab-instance of the debian-project (salsa.debian.org) where if can profit from their excellent CI-configurations.

sezanzeb:
I can't find your comment on splitting the packaging from the source code

Blaimi:
oh, sorry. it was this one: https://github.com/sezanzeb/key-mapper/pull/82#issuecomment-816592203. That's the difference between `3.0 (native)` and `3.0 (quilt)` in the file `debian/source/format`.

Of course you can. The benefits of salsa it that the package undergoes different tests automatically. like installations on different distributions and upgrades from old packages, etc.

But these discussions are not yes necessary. Ubuntu 21.04 is in beta, fedora 34 is in beta, debian 11 is in hard-freeze. I.e. the next deadline for upstream-packaging is at least 3 month ago. In the meantime we can try getting it into debian sid which migrates to testing which migrates to ubuntu, maybe set up a ppa and – if I get the fedora-rpm working – set up a copr (that's kind of ppa-equivalent in the fedora-world)

sezanzeb:
I'm really glad there are people who know what all of that stuff is and how it works. Because I don't.

I didn't know there is a deadline, I'd love to meet it then.

[...]

PR #351, opened 2022-03-22
Added installation guide for `fedora`

pourmand1376:
I had a lot of trouble installing this module on `fedora 35`. I couldn't install `evdev` using `sudo pip install evdev` due to some weird error. I ended up installing finding and installing these python packages and after installing them, your solution works just fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,16 @@ sudo apt install ./dist/input-remapper-1.4.2.deb
 
 input-remapper is now part of [Debian Unstable](https://packages.debian.org/sid/input-remapper)
 
+##### Fedora
+
+On fedora, you should first install python dependecies via below command and then continue with pip installation!
+
+```bash
+sudo dnf install python3-evdev python3-devel gtksourceview4 python3-pydantic
+sudo dnf install python-pydbus
+sudo dnf install xmodmap
+```
+
 ##### pip
 
 Dependencies: `python3-evdev` ≥1.3.0, `gtksourceview4`, `python3-devel`, `python3-pydantic`


README.md
```diff
@@ -36,6 +36,16 @@ sudo apt install ./dist/input-remapper-1.4.2.deb
...
 
+##### Fedora
+
+On fedora, you should first install python dependecies via below command and then continue with pip installation!
+
+```bash
+sudo dnf install python3-evdev python3-devel gtksourceview4 python3-pydantic
+sudo dnf install python-pydbus
+sudo dnf install xmodmap
+```
```
sezanzeb, line 47: - python3-evdev from the fedora repo is too old
- python-pydbus does not seem to exist: https://packages.fedoraproject.org/search?query=python-pydbus, did you mean python3-pydbus?

Except for python3-pydbus and xmodmap, which is an optional dependency, all of their exact names are already named in the pip section, so I'm not sure what the benefit of this would be.

I'd like to have a script here that generates a .rpm package file, so that the fedora package manager can properly manage it.

sezanzeb:
There is also this: https://copr.fedorainfracloud.org/coprs/sunwire/input-remapper/

README.md
pourmand1376: Theoretically, using `pip` should be the same. But, in the case of `fedora`, it is not.

pourmand1376: I think, I should also change `python-evdev` to `python3-evdev` according to [this](https://github.com/sezanzeb/input-remapper/issues/83#issuecomment-1030859666). But, as both of them are old as discussed, we eventually need to run `sudo pip install evdev`.

gvalkov:
Hello. python-evdev 1.5.0 is now available in fedora 35, 36 and rawhide. Hope this helps.

jason-bxtn:
For what it's worth, I just successfully installed on Fedora 37 following the manual install instructions in README.md, with one addition of `sudo dnf install gettext`, as I was getting a "no such file or directory" error when setup.py was trying to run `msgfmt` to generate translations.

pauliuszaleckas:
`sudo dnf install input-remapper` was all I needed on Fedora 38

richardm1:
> `sudo dnf install input-remapper` was all I needed on Fedora 38

Still works in Fedora 40.

Heinrich-XIAO (approved):

sezanzeb:
Since the readme already mentions `sudo dnf install input-remapper`, I guess this pull request should be closed, right?

pourmand1376:
Yep. It should be closed!

PR #369, opened 2022-04-14
Add `RequiresMountsFor=/var/log` to systemd unit

rocketraman:
If anyone mounts `/var` from a separate filesystem, this will ensure that `/var/log/input-remapper-control` is not written before `/var`/ is mounted.

### DIFF
--- a/data/input-remapper.service
+++ b/data/input-remapper.service
@@ -3,6 +3,7 @@ Description=Service to inject keycodes without the GUI application
 # dbus is required for ipc between gui and input-remapper-control
 Requires=dbus.service
 After=dbus.service
+RequiresMountsFor=/var/log
 
 [Service]
 Type=dbus


sezanzeb:
Thanks

rocketraman:
@sezanzeb For some reason, this still isn't working for me. I've confirmed that systemd is supposed to be mounting var before starting the `input-remapper` service:

```
# systemd-analyze critical-chain input-remapper.service
The time when unit became active or started is printed after the "@" character.
The time the unit took to start is printed after the "+" character.

input-remapper.service +468ms
└─basic.target @5.960s
  └─dbus-broker.service @5.913s +46ms
    └─dbus.socket @5.906s
      └─sysinit.target @5.903s
        └─systemd-update-utmp.service @5.895s +8ms
          └─auditd.service @5.819s +73ms
            └─systemd-tmpfiles-setup.service @4.901s +914ms
              └─systemd-journal-flush.service @3.704s +1.194s
                └─var.mount @3.579s +27ms
                  └─systemd-fsck@dev-mapper-vg_raid10\x2dlv_var.service @3.447s +130ms
                    └─dev-mapper-vg_raid10\x2dlv_var.device @3.445s
```

So from this it is clear that `input-remapper` should have been started much after `/var/` was mounted. However, after every reboot I still see that `/var/log/input-remapper-control` exists on the root filesystem, underneath the mounted `/var` filesystem.

Any ideas as to why that might be?

rocketraman:
Ah, I think I see the problem... the Fedora package I am using (from https://copr.fedorainfracloud.org/coprs/sunwire/input-remapper/) contains `/usr/lib/udev/rules.d/99-input-remapper.rules`, which runs `/bin/input-remapper-control` very early in the boot process.

Could `input-remapper-control` be initialized via systemd rather than udev? Or alternatively could there be a way to turn off or delay logging?

sezanzeb:
There is some logic that checks if systemd has finished booting before doing anything: https://github.com/sezanzeb/input-remapper/blob/main/bin/input-remapper-control#L181

If all log calls are removed until after https://github.com/sezanzeb/input-remapper/blob/main/bin/input-remapper-control#L221it would probably work.

A probably better solution would be to turn off logging to the file when started via udev, because logs are available in journalctl anyway. Same probably goes for the service

sezanzeb:
Getting logs via journalctl requires

```
sudo udevadm control --log-priority=debug
sudo udevadm control --reload-rules
```

beforehand though

sezanzeb:
Or apparently to write `udev_log="debug" in /etc/udev/udev.conf` (see https://blog.dachary.org/2016/11/30/logging-udev-events-at-boot-time/)

As long as there is some way to get logs when people have problems I'm fine with removing logfiles.

rocketraman:
Thank you for the comprehensive investigation @sezanzeb . Having the logs written to journalctl seems like the right way to go.

sezanzeb:
are you able to make another PR for this by any chance? I think there is an `add_filehandler` function or something, maybe just remove all usages of it, the function itself and the unittest.

rocketraman:
@sezanzeb I can give it a shot over the next few weeks.

sezanzeb:
> @sezanzeb I can give it a shot over the next few weeks.

thanks!

PR #375, opened 2022-04-20
Frontend Refactor

jonasBoss:


jonasBoss:
All the changes to the overall architecture of the UI are done. The UI is usable for managing presets, creating mappings and macros. Basically all the original features with the exception of mapping to axis are available in the UI.

@sezanzeb The UX is now largely tested with unittests. I am currently looking into testing the UI, do you know why we could not test the UI with unittests?
Also if you have time we can go over the changes I made, so I can incorporate any feedback.

sezanzeb:
What do you mean with UX? I know that as "user experience"

Sure, we can go over them soon. I'll try the changes out later

inputremapper/gui/data_bus.py
```diff
@@ -0,0 +1,171 @@
...
+
+# useful type aliases
+MessageListener = Callable[[Message], None]
+Capabilities = Dict[int, List]
+Name = str
+Key = str
+DeviceTypes = List[str]
+
+
+class DataBus:
```
sezanzeb, line 57: In which scenario would _messages contain more than one message?

I know this kind of service as a MessageBroker, like mosquitto or kafka

inputremapper/gui/controller.py
```diff
@@ -0,0 +1,317 @@
...
+from .data_bus import DataBus, MessageType, PresetData, StatusData, CombinationRecorded
+from .gettext import _
+from .data_manager import DataManager, DEFAULT_PRESET_NAME
+from .helper import is_helper_running
+from ..configs.mapping import MappingData, UIMapping
+from ..event_combination import EventCombination
+from ..injection.injector import RUNNING, FAILED, NO_GRAB, UPGRADE_EVDEV
+from ..input_event import InputEvent
+from ..logger import logger
+from .utils import CTX_SAVE, CTX_APPLY, CTX_KEYCODE, CTX_ERROR, CTX_WARNING, CTX_MAPPING
```
sezanzeb, line 36: What are your thoughts on relative vs absolute imports? I think other projects usually use absolute imports.

inputremapper/gui/data_manager.py
```diff
@@ -0,0 +1,446 @@
...
+DEFAULT_PRESET_NAME = "new preset"
+
+# useful type aliases
+Name = str
+GroupKey = str
+
+
+class DataManager:
+    """provides an interface to create configurations
+    and send commands to the service"""
```
sezanzeb, line 57: Can/shouldn't this be separated into one class for the config, and one for the communication?

inputremapper/gui/data_manager.py
```diff
@@ -0,0 +1,446 @@
...
+        Should send "groups" message to DataBus this will not happen immediately
+        because the system might take a bit until the groups are available
+        """
+        self._reader.refresh_groups()
+
+    def start_combination_recording(self):
+        """recorde user input
+        Will send "combination_recorded" messages as new input arrives.
+        Will eventually send a "recording_finished" message.
+        """
```
sezanzeb, line 430:
```suggestion
        """Record user input.
        
        Will send "combination_recorded" messages as new input arrives.
        Will eventually send a "recording_finished" message.
        """
```

inputremapper/gui/reader.py
```diff
@@ -61,192 +63,131 @@ class Reader:
...
+        """generator which yields live updated EventCombination's
 
-        self.previous_event = previous_event
-
-        if len(self._unreleased) > 0:
-            result = EventCombination(self._unreleased.values())
-            if result == self.previous_result:
-                # don't return the same stuff twice
-                return None
+        the generator will stop once all keys or inputs are released
```
sezanzeb, line 119: I think this docstring belongs to `_recorder`, not start_recorder`

inputremapper/gui/data_manager.py
```diff
@@ -0,0 +1,446 @@
...
+        self._reader.start_recorder()
```
sezanzeb, line 431: is it a reader or a recorder?

sezanzeb (commented):
Here are a few things I noticed

inputremapper/gui/data_bus.py
jonasBoss: when multiple listeners send messages they will be added to the `_messages` queue and sent once the current message is sent to the listeners.

inputremapper/gui/controller.py
jonasBoss: Not sure. I was testing this out just to see if it is more readable. I don't really have a preference, other than that we should keep it consistent everywhere. 

What do you think?

inputremapper/gui/data_manager.py
jonasBoss: Kinda, but not really. At first I had it separated into 3 classes: config, reader and daemon but I think that is not a good idea because I might end up integrating the helper in the daemon (the helper already uses the new event pipeline to read events).

I tried to combine reader and daemon in a communication class. 
But this communication class would be just a thin wrapper around the reader and the daemon without any logic itself.

The config needs to know about the reader in order to find all groups, the daemon needs to know about the config, so it can start injecting the correct preset. So maybe we can have them separated but they would still share each other as a object.

Having them in on class has the advantage that all the state of the application is in one place so it can act as a interface if someone else decides to create a UI.

inputremapper/gui/reader.py
jonasBoss: You are right. And the docstring needs updating the genarator actually yields nothing. It sends messages on the data_bus

inputremapper/gui/data_manager.py
jonasBoss: The class is a reader which reads all messages from the helper.
It has a recorder which can be started to record User input.

we can also call the whole thing a recorder or reader. whatever you prefer.

jonasBoss:
> What do you mean with UX? I know that as "user experience"

I mean the behavior of the application essentially everything the Controller and the DataManager does can be tested with unittests.

inputremapper/gui/data_bus.py
sezanzeb: but every `send` calls `_send_all` right away

inputremapper/gui/controller.py
sezanzeb: Relative imports have some special rules, absolute imports are easier to understand. So I'd go with absolute

inputremapper/gui/data_manager.py
sezanzeb: > integrating the helper in the daemon

the reason why they are separate is, because no user process should be able to ask the daemon for input events at all times. Only during the lifetime of the gui, which is also the lifetime of the helper, is reading keys permitted for the user processes.

Maybe there is an alternative solution with a pipe that is only readable from by elevated processes. But since the gui can't run as root in wayland a helper process would still be needed

inputremapper/gui/data_manager.py
sezanzeb: what about `HelperClient`?

sezanzeb:
> I am currently looking into testing the UI, do you know why we could not test the UI with unittests?

You can make the GTK widgets emit click events and such. Other than that, I don't know if it is possible to do anything more sophisticated with simulating a mouse and stuff.

inputremapper/gui/data_bus.py
jonasBoss: and `_send_all` returns immediately if it is already running. See `TestDataBus.test_preserves_order` this test will queue up 3 messages.

The idea is that all messages are always sent in the order in which they arrive. For example `Controller.on_preset_changed()` listens for a `preset` message and immediately responds by sending a `mapping` message. But we don't want this message to be sent immediately, first the `preset` message must be distributed to all listeners. 
Otherwise a listener who listens to both the `preset` and the `mapping` might receive the `mapping` first and mess up something.

Sure, we can rename it to message broker if this is a more common name.

inputremapper/gui/controller.py
jonasBoss: didn't know about that. Will change it to absolute.

inputremapper/gui/data_manager.py
jonasBoss: yeah this is nothing I will do anytime soon. And the daemon is currently not designed to do this easily. But it would open a lot of possibilities.

inputremapper/gui/data_bus.py
sezanzeb: Are multiple threads interacting with the DataBus?

jonasBoss: No.

sezanzeb: While the loop below is running, how would the python process ever have a chance of calling `send`?

since `_send` is not async and there are no other threads, the process will block until the loop is done.

Ah, except if the listener calls `send` as well. A second listener would then get the two messages in reverse order.

jonasBoss: > Ah, except if the listener calls `send` as well. A second listener would then get the two messages in reverse order.

exactly `TestDataBus.test_preserves_order` demonstrates this

sezanzeb:
> Also if you have time we can go over the changes I made, so I can incorporate any feedback.

I'll get back to you soon. I have a busy week ahead

sezanzeb:
I want to try to implement a new mapping handler at some point so that I have a good understanding of the beta architecture. Any suggestion on what would be a good one to have? This one maybe: https://github.com/sezanzeb/input-remapper/issues/380?

jonasBoss:
> I want to implement a new mapping handler at some point so that I have a better understanding of the beta architecture. Any suggestion on what would be a good one to have? This one maybe: #380?

Yeah, that is a good candidate. Should be rather simple and I think we have multiple issues which would be solved by this.

jonasBoss:
Since the new architecture for the Gui is done and test are all passing again I started work on implementing GUI solutions for all the new Features.
I appreciate any feedback and ideas how we can best implement the different features form a UI/UX perspective.

For now I have implemented the ability to give each mapping a individual name:
image

jonasBoss:
It is Finally done!!
All the most important settings are in the UI. It is time to test this more in-depth and generally gather feedback.
image
image

While there are still some things to do such as creating more integration tests and reworking all the documentation. I think we should already merge this into `beta` because it provides a way better UI than the one which is currently in `beta`. It also has some bug fixes to the service.

sezanzeb:
- draw lines in ui from border to border
"remove this input" error message when key exists
- output axis while macro code in text input: human readable error
- StatusData docstring
- _private underscores in components.py
- analog-axis not tested yet
- [x] "event" messsagebroker message to "selected-event"
- replacing `eval` with some parsing
- rename callbacks in context to notify_callbacks
- maybe move event_header.py out of the injection folder, since helper uses it as well.
- document why mapping_handlers are an improvement over the evdev.InputEvent directly from the read_loop in the helper

Here are the paddings/margins in px. I couldn't find it in the current docs anymore, but it was on archive.org:

image

jonasBoss:
From my side this is ready to merge. I implemented the changes we discussed in the review.

sezanzeb:
I'll be away in August. Would you like to be maintainer?

jonasBoss:
Yes, I think that makes sense.

jonasBoss:
When do you think your are able to implement the RelToRelHandler?
I think I will do the a RelToAbs an AbsToAbs in the next days. As soon as we have the missing handlers it would be a good point to finish up the beta branch and create a new release.

sezanzeb:
> When do you think your are able to implement the RelToRelHandler?

I honestly don't know if I can start working on it until September

tests/unit/test_event_pipeline/test_mapping_handlers.py
```diff
@@ -0,0 +1,261 @@
...
+            global_uinputs.get_uinput("mouse").write_history
+        )
+        self.assertIn(InputEvent.from_tuple((EV_KEY, BTN_LEFT, 0)), history[-2:])
+        self.assertIn(InputEvent.from_tuple((EV_KEY, BTN_RIGHT, 0)), history[-2:])
+        self.assertEqual(len(history), 4)
+
+
+class TestRelToBtnHanlder(BaseTests, unittest.IsolatedAsyncioTestCase):
+    def setUp(self):
+        self.handler = AxisSwitchHandler(
```
sezanzeb, line 254: this creates an AxisSwitchHandler in TestRelToBtnHanlder

jonasBoss: oh, that looks like an error. I was probably to lazy with the copy paste.

PR #395, opened 2022-05-07
Revert "Macro syntax check when typing"

sezanzeb:
Reverts sezanzeb/input-remapper#394

because I screwed the commit message up, sorry. missing kearkyle as contributor


sezanzeb:
kearkyle is still missing in the contributors list. Sorry

Why, github, why is this so complicated.

PR #397, opened 2022-05-08
Improved macro exception handling, keystroke sleep cleanup

sezanzeb:
- the sleep between mouse movement events is something different and should be constant
- removed redundant `_keycode_pause`
- when a child macro fails, the whole macro will stop now. This was needed because certain broken macros caused fast infinite loops that froze input-remapper.

### DIFF
--- a/inputremapper/injection/macros/macro.py
+++ b/inputremapper/injection/macros/macro.py
@@ -496,8 +496,8 @@ class Macro:
         async def task(handler):
             resolved_speed = value * _resolve(speed, [int])
             while self.is_holding():
-                await asyncio.sleep(injection_throttle / 1000)
                 handler(EV_REL, code, resolved_speed)
+                await asyncio.sleep(injection_throttle / 1000)
 
         self.tasks.append(task)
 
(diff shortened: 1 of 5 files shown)


sezanzeb:
@jonasBoss there is a new test for the changes to the exception handling. It basically just lets the exception of any child macro bubble up and the code that executed `macro.run(...)` has to handle it. I hope it is not too much work to incorporate that into beta.

jonasBoss:
should be no problem, once I merge main into beta again I will modify the MacroHandler accordingly

inputremapper/injection/macros/macro.py
```diff
@@ -492,11 +490,14 @@ def add_mouse(self, direction, speed):
...
 
+        # how long to pause in ms between the injection of mouse events
+        injection_throttle = 10
+
         async def task(handler):
             resolved_speed = value * _resolve(speed, [int])
             while self.is_holding():
                 handler(EV_REL, code, resolved_speed)
-                await self._keycode_pause()
+                await asyncio.sleep(injection_throttle / 1000)
```
jonasBoss, line 500: I will use `await asyncio.sleep(1 / self.mapping.rate)` here, similar to `add_wheel`
That way the event repetition rate is can be controlled by the mapping and the `Mapping.rate` is meant to do that for all `EV_REL` events

sezanzeb: ok

PR #399, opened 2022-05-10
Merge branch 'main' into beta

jonasBoss:


sezanzeb:
It seems `test_hold_failing_child` is missing

jonasBoss:
I see, that was part of TestKeycodeMapper, which no longer exits. That explains why I missed it during the merge. I will implement it in TestMacros.
Thanks for double checking.

PR #418, opened 2022-06-11
Removed logging filehandlers

sezanzeb:
- Journalctl should be used instead, see https://github.com/sezanzeb/input-remapper/pull/369#issuecomment-1107799124
- I never needed people to post contents from logfiles to fix problems

To get logs via journalctl during boot, write `udev_log="debug"` into `/etc/udev/udev.conf` (see https://blog.dachary.org/2016/11/30/logging-udev-events-at-boot-time/, https://askubuntu.com/questions/1190823/how-to-see-all-udev-events-during-boot)


Techwolf:
My system, gentoo, and many others do NOT have journalctl. And for many reasons, most are good ones.

PR #440, opened 2022-07-23
AbsToAbsHandler

jonasBoss:
only needs some tests and then should be good to merge

### DIFF
--- a/inputremapper/configs/mapping.py
+++ b/inputremapper/configs/mapping.py
@@ -292,15 +292,24 @@ class Mapping(BaseModel):
         return values
 
     @root_validator
-    def output_axis_given(cls, values):
-        """Validate that an output type is an axis if we have an input axis."""
+    def output_matches_input(cls, values):
+        """Validate that an output type is an axis if we have an input axis.
+        And vice versa"""
         combination: EventCombination = values.get("event_combination")
         output_type = values.get("output_type")
         event_values = [event.value for event in combination]
-        if 0 not in event_values:
-            return values
+        output_symbol = values.get("output_symbol")
 
-        if output_type not in (EV_ABS, EV_REL):
+        if 0 not in event_values and not (
+            output_symbol is not None or output_type == EV_KEY
+        ):
+            raise ValueError(
+                f"missing macro or key: "
+                f"the {combination = } specifies a key input, "
+                f"but no output macro or key is programmed"
+            )
+
+        if 0 in event_values and output_type not in (EV_ABS, EV_REL):
             raise ValueError(
                 f"missing output axis: "
                 f"the {combination = } specifies a input axis, "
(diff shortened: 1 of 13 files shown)


jonasBoss:
This fixes #441 and implements the AbsToAbsHandler.
From my side it is ready to merge.

inputremapper/configs/mapping.py
```diff
@@ -292,15 +292,24 @@ def validate_output_integrity(cls, values):
...
         output_type = values.get("output_type")
         event_values = [event.value for event in combination]
-        if 0 not in event_values:
-            return values
+        output_symbol = values.get("output_symbol")
 
-        if output_type not in (EV_ABS, EV_REL):
+        if 0 not in event_values and not (
+            output_symbol is not None or output_type == EV_KEY
+        ):
```
sezanzeb, line 305: I think this is equivalent:

```suggestion
        if 0 not in event_values and output_symbol and output_type != EV_KEY:
```

inputremapper/injection/mapping_handlers/abs_to_abs_handler.py
```diff
@@ -0,0 +1,153 @@
...
+            return int(max(self._target_absinfo.min, y))
+
+    def _write(self, value: int):
+        """Inject."""
+        try:
+            global_uinputs.write(
+                (*self._output_axis, value), self.mapping.target_uinput
+            )
+        except OverflowError:
+            # screwed up the calculation of mouse movements
```
sezanzeb, line 141: mouse movements?

inputremapper/injection/mapping_handlers/abs_to_abs_handler.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def __init__(
+        self,
+        combination: EventCombination,
+        mapping: Mapping,
+        **_,
+    ) -> None:
+        super(AbsToAbsHandler, self).__init__(combination, mapping)
+
+        # find the input event we are supposed to map
```
sezanzeb, line 55: I think I'd add an example to this. I hope I got this right

```suggestion
        # find the input event we are supposed to map. If the input combination is
        # BTN_A + ABS_X + BTN_B, then use the value of ABS_X for the transformation
```

inputremapper/injection/mapping_handlers/abs_to_abs_handler.py
jonasBoss: copy paste error 
```suggestion
            # screwed up the calculation the event value
```

inputremapper/configs/mapping.py
jonasBoss: almost, this works:
```suggestion
        if 0 not in event_values and not output_symbol and output_type != EV_KEY:
```

sezanzeb:
Thanks

PR #442, opened 2022-07-25
RelToAbs

jonasBoss:
needs tests

### DIFF
--- a/data/input-remapper.glade
+++ b/data/input-remapper.glade
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.38.2 -->
+<!-- Generated with glade 3.40.0 -->
 <interface>
   <requires lib="gtk+" version="3.22"/>
   <requires lib="gtksourceview" version="4.0"/>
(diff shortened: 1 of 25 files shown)


jonasBoss:
This should be a good example for a `RelToRelHandler`

sezanzeb:
image

Is this related to this PR?

sezanzeb:
ah, I had to clear the text input. Now it says:

image

jonasBoss:
Those have nothing to do with this PR. I created #468 for this, there we can collect all the problematic error messages and fix them.

sezanzeb:
what should I do about the second error message in order to try the RelToAbs handler out?

jonasBoss:
You need to configure the input as axis input. Go to the `Advanced Input Configuration` and set the flip the swich "Use as Analog"

sezanzeb:
If no input is used as analog, but the mapping maps to "Analog Axis", can we default to using the first analog input as analog axis?

I think user experience is really one of the most important things in software. If we can somehow make it easier for people to use it I'm all for it.

jonasBoss:
I can see what I can come up with.

jonasBoss:
I just updated the GUI. It will now show a message when the user switches between "Key or Macro" and "Analog axis" and it will remove the macro input or the axis input to avoid errors. 
It is not perfect, the user still needs to update the input accordingly but hopefully it clears up some confusion. 

This needs some integration tests.

jonasBoss:
@sezanzeb can you have another look at this? Should be ready.

sezanzeb:
great! Thanks. It was easy to set it up

Can you please add a tooltip to "Output Axis" that says something like "Available output axes are affected by the Target setting."? I'll merge it then

sezanzeb:
I just started working on the new UI, I'll add the tooltip later

PR #458, opened 2022-08-24
Improved Test Fixtures

jonasBoss:
This allows us to specify the exact device to which a InputEvent is sent.
Previously it was only possible to specify the Group, which was limiting for some tests.

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -23,8 +23,10 @@
 
 This module needs to be imported first in test files.
 """
+from __future__ import annotations
 
 import argparse
+import dataclasses
 import json
 import os
 import sys
@@ -32,11 +34,19 @@ import tempfile
 import traceback
 import warnings
 from multiprocessing.connection import Connection
-from typing import Dict, Tuple
+from typing import Dict, Tuple, Optional
 import tracemalloc
 
 tracemalloc.start()
 
+# ensure nothing has loaded
+if module := sys.modules.get("inputremapper"):
+    imported = [m for m in module.__dict__ if not m.startswith("__")]
+    raise AssertionError(
+        f"the modules {imported} from inputremapper where already imported, this can "
+        f"cause issues with the tests. Make sure to always import tests.test before any"
+        f" inputremapper module"
+    )
 try:
     sys.modules.get("tests.test").main
     raise AssertionError(
@@ -151,7 +161,7 @@ tmp = temporary_directory.name
 uinput_write_history = []
 # for tests that makes the injector create its processes
 uinput_write_history_pipe = multiprocessing.Pipe()
-pending_events: Dict[str, Tuple[Connection, Connection]] = {}
+pending_events: Dict[Fixture, Tuple[Connection, Connection]] = {}
 
 
 def read_write_history_pipe():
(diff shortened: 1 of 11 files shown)


tests/test.py
```diff
@@ -262,79 +293,190 @@ def read_write_history_pipe():
...
+            self.dev_input_event15,
+            self.dev_input_event20,
+            self.dev_input_event30,
+            self.dev_input_event31,
+            self.dev_input_event40,
+            self.dev_input_event51,
+        ]
+        self._dynamic_fixtures = {}
+
+    def __getitem__(self, item: str) -> Fixture:
```
sezanzeb, line 350: Please add a docstring that explains `item`

tests/test.py
```diff
@@ -170,21 +180,36 @@ def read_write_history_pipe():
...
+    name: str = "unset"
+    info: evdev.device.DeviceInfo = evdev.device.DeviceInfo(None, None, None, None)
+    phys: str = "unset"
+    group_key: Optional[str] = None
+
+    def __hash__(self):
+        return hash(self.path)
+
+
+class _Fixtures:
```
sezanzeb, line 197: Docstring please.

It seems `_Fixtures` is some sort of container. Is it like a dict?

tests/test.py
```diff
@@ -262,79 +293,190 @@ def read_write_history_pipe():
...
+    @property
+    def foo_device_2_keyboard(self):
+        return self["/dev/input/event10"]
+
+    @property
+    def foo_device_2_13(self):
+        return self["/dev/input/event13"]
+
+    @property
+    def foo_device_2_qux(self):
```
sezanzeb, line 403: is `self.foo_device_2_qux` the same as `self.dev_input_event14`?

tests/test.py
```diff
@@ -823,13 +963,13 @@ def prepare_presets():
...
 
-    time.sleep(0.05)
+    time.sleep(0.1)
     preset2 = Preset(get_preset_path("Foo Device", "preset2"))
     preset2.add(get_key_mapping(combination="1,3,1"))
     preset2.add(get_key_mapping(combination="1,4,1"))
     preset2.save()
 
-    time.sleep(0.05)  # make sure the timestamp of preset 3 is the newest
+    time.sleep(0.1)  # make sure the timestamp of preset 3 is the newest
```
sezanzeb, line 972:
```suggestion
    # make sure the timestamp of preset 3 is the newest
    # (probably for the GUI to load this preset first)
    time.sleep(0.1)
```

sezanzeb (changes requested):
Very nice. It improves the readability of tests a lot

tests/test.py
jonasBoss: It is just a class with a bunch of class variables.
It defines all the standard fixtures and provides access to them via the path (`fixtures["/dev/input/eventX"]`) the attribute name (`fixtures.dev_input_eventX`) and a property name (`fixtures.foo_device_2_mouse`).
It also allows dynamic creation of new fixtures during runtime.

tests/test.py
jonasBoss: yes

sezanzeb: isn't it redundant then?

jonasBoss: yes it is. I created the property for convenience to improve the readability. The attribute name `dev_input_eventXX` allows us to access the attribute with `fixtures["/dev/input/eventXX"]` from the `__getitem__` method, which is necessary for backwards comparability with some tests and for dynamic fixtures.

sezanzeb: ok

Issue #473, opened 2022-09-09
GUI Redesign based on libadwaita and multiple "pages"


sezanzeb:
I'll try to collect ideas and resources for a new GUI here. I think it is really hard to fit all of the features of the beta branch into one single window.

goals:
- a clean and non-cluttered GUI, similar to gnome-music
- work well with the modern adwaita theme
- use headerbars (maybe for a pagination, and to go back in the menu)
- reuse as many elements from the mapping configuration of the beta GUI as possible / recreate them with libadwaita
- use modern dialogs https://developer.gnome.org/hig/patterns/feedback/dialogs.html
- respect hig spacings https://developer-old.gnome.org/hig/stable/visual-layout.html.en

concept:
1. application starts and shows a device list. Analogous to how gnome-music shows the album list on start
2. select device from list. The complete GUI transitions to a different view, which shows the preset list. gnome-music also has that transition. On the top left in the headerbar a back button `<` is shown
3. selecting a preset from list acts in the same way
4. This leads to the preset configuration, which might look exactly like the stuff we have on beta

resources:
- adwaita python demo https://github.com/timlau/adw_pydemo
- gnome-music, which is also written in python https://gitlab.gnome.org/GNOME/gnome-music/
- libadwaita docs https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/
- cambalache, the GUI designer for libadwaita/gtk4 https://www.linuxadictos.com/en/swap-a-tool-for-creating-gtk-user-interfaces.html

sezanzeb:
They do things like

```
@Gtk.Template(resource_path="/org/gnome/Music/ui/AlbumsView.ui")
class AlbumsView(Gtk.Stack):
```

jonasBoss:
I also think it is a good idea to have a wizard with a more curated process for creating a new mapping.

1. What kind of Mapping?
    - Analog axis
    - key
    - macro
2. Record the input
    - provide a warning if the input is incompatible with the selected mapping type
3. provide a window with the relevant information for mapping type

TerrorSquad:
How about instead of using libadwaita, you port the app to Tauri and write the frontend in something like Vue.js?

sezanzeb:
I want desktop apps to properly integrate with the looks of the rest of the desktop

TerrorSquad:
Understandable.

But that's only for the GNOME desktop? Is it going be easy to integrate with the look and feel of KDE or Cinnamon?

[...]

PR #484, opened 2022-10-02
UI 3

sezanzeb:
closes https://github.com/sezanzeb/input-remapper/issues/473

will be GTK3 based, because I don't like the theming issues in libadwaita, and because I'd like to save time by continuing to use the GTK version we are using now

migrating to GTK4 might be interesting, because I think then the bottom corners will be rounded as well in gnome


sezanzeb:
image

sezanzeb:
image

sezanzeb:
image

sezanzeb:
image

sezanzeb:
tests missing

sezanzeb:
When hitting the "Key or Macro | Analog Axis" buttons without an existing Mapping, it would freeze them into the active state.

`NoMappingError` exists now to revert the button state when this happens

sezanzeb:
image

sezanzeb:
image

sezanzeb:
todo: disable the output box completely if no mapping is selected

jonasBoss:
I didn't test this yet, but I like the use of multiple pages. This makes for a much cleaner UI. I will have more time to look at this in one or two weeks.

I feel like the 3-page version is better with a few changes:
  - The presets page should allow for renaming/copying/applying... presets
  - Rename the Editor page to Mappings because it no longer has preset related stuff on it

sezanzeb:
Thanks.

> The presets page should allow for renaming/copying/applying... presets

In that case I would do that by adding buttons to the boxes

image

I want to be finished with this and the Rel2Rel mapping within the next week though. I'll think about it. We can still modify the GUI later on again and reuse the new components for your suggestion.

I'll try to make a base class for those boxes and will try to keep in mind that we might eventually add buttons to them.

jonasBoss:
> In that case I would do that by adding buttons to the boxes

Sounds like a good idea.

> We can still modify the GUI later on

That is right, I like the changes you made so far and don't see any issue merging them into beta.

I will look into the GUI when I have more time.

tests/integration/test_components.py
```diff
@@ -1024,7 +1059,7 @@ class TestKeyAxisStackSwitcher(ComponentBaseTest):
     def setUp(self) -> None:
         super(TestKeyAxisStackSwitcher, self).setUp()
         self.gui = Gtk.Box()
-        self.gtk_stack = Gtk.Stack()
+        self.gtk_stack = inputremapper.gui.components.main.Stack()
```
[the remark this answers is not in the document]
sezanzeb: automatic refactoring ftw

sezanzeb:
You can use this link to review changes up to the components.py split: https://github.com/sezanzeb/input-remapper/pull/484/files/af8fd461846d8e392e1d4515d00a04ed2ebede07..b562642ff48372cab7b51a57f66adac0b7edc714

Otherwise the diff includes moved classes that didn't change

sezanzeb:
@jonasBoss do you remember why `Gtk.main()` is being called in `ComponentBaseTest.tearDown`?

sezanzeb:
allright, so `TestRecordingToggle` causes all the freezes, because it is `self.toggle_button` instead of `self.gui`

I can't find any logic at all requiring widgets to be stored in `self.gui`

magic

jonasBoss:
> @jonasBoss do you remember why `Gtk.main()` is being called in `ComponentBaseTest.tearDown`?

`Gtk.main()` will start the Gtk event loop and process all pending events. So the gui will do whatever is queued up this ensures that the next tests starts without pending events. 
We are also calling:
```python
GLib.timeout_add(0, self.gui.destroy)
GLib.timeout_add(0, Gtk.main_quit)
```
just before the `Gtk.main()` so the gui will properly shutdown.

> I can't find any logic at all requiring widgets to be stored in self.gui

yeah it is not very obvious, but there is a docstring in `ComponentBaseTest`:
```python
class ComponentBaseTest(unittest.TestCase):
    """test a gui component

    ensures to tearDown self.gui
    all gtk objects must be a child of self.gui in order to ensure proper cleanup"""
```

jonasBoss:
I get a black bar near the line numbers when the editor becomes so long that it has a scroll bar:
image
image

tests/integration/test_user_interface.py
```diff
@@ -104,4 +104,6 @@ def test_combination_label_shows_text_when_empty_mapping(self):
         gtk_iteration()
         label: Gtk.Label = self.user_interface.get("combination-label")
         self.assertEqual(label.get_text(), "no input configured")
-        self.assertEqual(label.get_opacity(), 0.4)
+
+        # 0.5 != 0.501960..., for whatever reason this number is all screwed up
+        self.assertLess(abs(label.get_opacity() - 0.5), 0.1)
```
jonasBoss, line 109:
```suggestion
        self.assertAlmostEqual(label.get_opacity(), 0.5, delta=0.1)
```

sezanzeb:
your new gui code is very nice to work with

sezanzeb:
image

image

image

inputremapper/gui/components/editor.py
```diff
@@ -942,6 +942,63 @@ def _on_gtk_toggle(self, btn: Gtk.ToggleButton):
             self._controller.update_mapping(mapping_type="analog")
 
 
+class Colors:
```
jonasBoss, line 945: This Probably belongs into `inputremapper.gui.utils`.

sezanzeb:
Here is the diff after the components.py split: https://github.com/sezanzeb/input-remapper/pull/484/files/9b94df472a6ef49b3012ec1fe88d59da651845b4..249255057e0b73acb19b9e8695f789e09f44d977

inputremapper/gui/components/common.py
```diff
@@ -0,0 +1,183 @@
...
+        self._group_key = data.group_key
+        self._render()
+
+    def _on_mapping_changed(self, mapping: MappingData):
+        if mapping.name:
+            self._mapping_name = mapping.name
+        elif mapping.event_combination != EventCombination.empty_combination():
+            self._mapping_name = mapping.event_combination.beautify()
+        else:
+            self._mapping_name = _("empty mapping")
```
jonasBoss, line 167:
```suggestion
    def _on_mapping_changed(self, mapping: MappingData):
        self._mapping_name = mapping.get_name()
```
Can we create a `Mapping.get_name()` function instead of resolving the correct name here?

There is also similar logic in the `gui.components.editor.MappingSelectionLabel.name` property. Which is no longer needed, as we can just expose the private `MappingSelectionLabel._name` attribute.

inputremapper/gui/components/editor.py
```diff
@@ -294,11 +199,14 @@ def __init__(
...
 
+        self.label.set_margin_top(11)
+        self.label.set_margin_bottom(11)
+
         # button to edit the name of the mapping
         self.edit_btn = Gtk.Button()
         self.edit_btn.set_relief(Gtk.ReliefStyle.NONE)
         self.edit_btn.set_image(
-            Gtk.Image.new_from_icon_name(Gtk.STOCK_EDIT, Gtk.IconSize.MENU)
+            Gtk.Image.new_from_icon_name("text-x-generic", Gtk.IconSize.MENU)
```
jonasBoss, line 209: I feel like the edit symbol is better at communicating the purpose of the button.

image
image

inputremapper/gui/components/editor.py
```diff
@@ -294,11 +199,14 @@ def __init__(
         # set the name or combination.beautify as label
         self.label.set_label(self.name)
 
+        self.label.set_margin_top(11)
+        self.label.set_margin_bottom(11)
+
         # button to edit the name of the mapping
         self.edit_btn = Gtk.Button()
         self.edit_btn.set_relief(Gtk.ReliefStyle.NONE)
```
jonasBoss, line 207:
```suggestion
```
this relief highlights the button better, although it is still difficult to see. Maybe there is a different theme accent color which we can use to improve the visibility of this button.
image

jonasBoss (approved):

inputremapper/gui/components/editor.py
sezanzeb: Stock icons seem to be deprecated (at least glade claims that it is). The `Gtk.STOCK_EDIT` icon isn't integrating well with other icons in GNOME.

But you are right, the STOCK_EDIT is still better

inputremapper/gui/components/editor.py
sezanzeb: I'll keep None since the stock icon is back now

inputremapper/gui/components/editor.py
jonasBoss:
```python
        self.edit_btn.set_image(
            Gtk.Image.new_from_icon_name("edit", Gtk.IconSize.MENU)
        )
```
seems to work just fine.

sezanzeb: not in gnome: image

inputremapper/gui/data_manager.py
```diff
@@ -195,13 +197,12 @@ def get_preset_names(self) -> Tuple[Name, ...]:
...
-    def get_mappings(self) -> Optional[List[Tuple[Optional[Name], EventCombination]]]:
+    def get_mappings(self) -> Optional[List[MappingData]]:
         """all mapping names and their combination from the active_preset"""
         if not self._active_preset:
             return None
-        return [
-            (mapping.name, mapping.event_combination) for mapping in self._active_preset
-        ]
+
+        return list(self._active_preset)
```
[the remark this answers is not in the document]
jonasBoss:
```suggestion
        return [mapping.get_bus_message() for mapping in self._active_preset]
```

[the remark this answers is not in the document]
sezanzeb: thanks

inputremapper/gui/components/common.py
jonasBoss: If the circular import is only because of the typing you can do this:

``` python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
   # imports that are not needed during runtime
```

sezanzeb: Since it works now I would probably just stick with the split file

sezanzeb: Is this ready to merge now?

jonasBoss: looks good.

PR #487, opened 2022-10-08
Added ru_RU translation and adjusted ru text for policykit file

yngmjgsd:
Hello,

I've added **ru_RU** translation for **input-mapper**. Tested on my system and it works, though some distros specify **LANGUAGE** variable in a way there is only one letter group for the used language (for example, only **ru** instead of **ru_RU**). It looks like this:

`user@localhost ~ $ echo $LANGUAGE
ru:en_US
user@localhost ~ $`

I suggest you do the following changes - adjust the installation (or build) script in a way it creates a folder with a locale inside it and a directory symlink (sorry, I don't know much about installation, or build scripts for this application so I can't suggest more practical changes), like this:

```
user@localhost /usr/share/input-remapper/lang $ sudo ln -s ru_RU ru
user@localhost /usr/share/input-remapper/lang $ ls -l
total 16
drwxr-xr-x 3 root root 4096 окт  8 02:57 it_IT
lrwxrwxrwx 1 root root    5 окт  9 00:47 ru -> ru_RU
drwxr-xr-x 3 root root 4096 окт  8 22:53 ru_RU
drwxr-xr-x 3 root root 4096 окт  8 02:57 sk_SK
drwxr-xr-x 3 root root 4096 окт  8 02:57 zh_CN
user@localhost /usr/share/input-remapper/lang $
```

So it will pick up both if necessary and display translation anyway. Looks like an easy solution.

Thanks for your work by the way. Was able to adjust my Logitech MX Vertical just fine. The translation file was tested and it works:

изображение

### DIFF
--- a/data/input-remapper.policy
+++ b/data/input-remapper.policy
@@ -8,6 +8,7 @@
         <description>Run Input Remapper as root</description>
         <message>Authentication is required to discover and read devices.</message>
 	<message xml:lang="sk">Vyžaduje sa prihlásenie na objavenie a prístup k zariadeniam.</message>
+	<message xml:lang="ru">Требуется аутентификация для обнаружения и чтения устройств.</message>
         <defaults>
             <allow_any>no</allow_any>
             <allow_inactive>auth_admin_keep</allow_inactive>
--- a/po/ru_RU.po
+++ b/po/ru_RU.po
@@ -1,0 +1,343 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: \n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2022-03-27 13:08+0200\n"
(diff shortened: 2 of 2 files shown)


sezanzeb:
Thanks!

Do you know which distros do that?

yngmjgsd:
> Thanks!
> 
> Do you know which distros do that?

I'm positive it can be almost any distro with the following settings:

1. Modern KDE Plasma 5 (like the one in Kubuntu 22.04)
2. More than one language in Localization settings
3. RU is the first one (en_US is the second in priority). Maybe there can be even a third one, but I don't know if it makes any difference. Maybe it affects some other non en_US languages.

For example, this is how it looks in the **systemsettings5** application:

изображение

And this is how the **LANGUAGE** variable is formed:

изображение

So, perhaps this is KDE software that is responsible for this environment setting. I've seen it on:

- Kubuntu 22.04
- ROSA Linux 2021.2 (KDE)
- Manjaro 22 (KDE)

Maybe there are more distros, but it looks like more of a KDE-specific thing. Making a symlink with a short locale name might resolve it on some distros/DEs I guess.

sezanzeb:
allright, thanks.

I'll create a new issue for that matter

sezanzeb:
I pushed a symlink from po/ru to po/ru_RU to the repo. Does it work in KDE now?

yngmjgsd:
> 

Yup. After re-installation, it picked the translation just fine. Thanks!

Issue #494, opened 2022-10-16
increment/decrement variable ("add" macro)


nelsam:
[note: I'm a developer myself, so I have a particular way of writing feature requests.  Feel free to change my formatting to match your own preferences :)]

Use case statement:

As an Elden Ring player trying to separate the "sprint" and "dodge" actions
I want to be able to increment and decrement variables
So that I can perform a key_up only after the last movement key is released.

Description:

I'm trying to get Elden Ring to act like it has separate bindings for "sprint" and "dodge".  Unfortunately, the way that the game handles dodging makes this ... _complicated_.  Long story short: if I use `key_down($sprint_key).wait(500).hold().key_up($sprint_key)`, it works perfectly when I'm already moving, but if I'm not moving then Elden Ring decides that I actually wanted to dodge.  This can be disastrous if I press sprint _before_ pressing a movement key when I'm trying to sprint away from a threat - I can get locked into the dodging animation.

My next attempt was to set a variable when sprint is pressed _and also_ to set a variable when any movement key is pressed.  Only when both variables are set do I call `key_down($sprint_key)`.

The problem with that is that multiple movement keys can be held down at the same time, so when I hold left, then left+up, then up, it detects that I let go of left and stops sprinting.

So what I really want on my movement keys is:

```
key_down($left_key).
  add(moving, 1).
  # only press the sprint key if this is the first movement key pressed *and* sprint is being held
  if_eq($moving, 1, if_eq($sprint, 1, key_down($sprint_key), None), None).
  hold().
  add(moving, -1).
  # only release the sprint key if this was the last movement key released *and* sprint is being held
  if_eq($moving, 0, if_eq($sprint, 1, key_up($sprint_key), None), None).
  key_up($left_key)
```

I can work around this for now by setting a different variable for each movement key, but it's _very_ ugly. That `if_eq` statement is much worse when I have to read four variables instead of one.

Acceptance:
1. Create a macro that increments a variable.
    - e.g. `add(foo, 1)`
2. Create a macro that only triggers when a variable has been incremented N times.
    - e.g. `if_eq($foo, 3, key(key_a), None)`
4. Ensure that the macro from [2] only triggers when the macro from [1] has run the correct number of times.

sezanzeb:
this just arrived on the beta branch

PR #496, opened 2022-10-17
`add` macro

sezanzeb:
closes https://github.com/sezanzeb/input-remapper/issues/494

---

EDIT: only added the "add" macro, and did some cleanup

macro syntax updates:

will use semicolons instead dots in the future, because in the past dots basically just told the parser that the next statement is following. Semicolons are the more natural choice for that.

as of now a statement-terminator symbol would not be required at all, because after the last closing bracket ")" the statement is always finished. But that might be a problem in the future if more sophisticated syntax is added. People are used to semicolons in other languages already, so it should be fine for everyone.

```c#
key(a);
key(b);
repeat(
    5,
    key(c);
    key(d);
);
```

I'm thinking about adding curly braces for the same reasons

```c#
key(a);
key(b);
repeat(5, {
    key(c);
    key(d);
});
```

By using more standard language features in the macro syntax, the predefined syntax highlighters of gtksourceview might also work better.


sezanzeb:
Due to uncertainty I'll only do the add macro

Issue #495, opened 2022-10-16
Automatically set "Use as analog" if output is analog


sezanzeb:
- if the output type is "Analog Axis"
- after recording finished
- if an analog input has been recorded

then set "Use as analog" automatically for the first analog input in the combination

sezanzeb:
If EV_REL input is recorded as analog axis, and if the deadzone is not overwritten manually already, default it to 0

PR #497, opened 2022-10-17
RelToRel

sezanzeb:


inputremapper/injection/mapping_handlers/rel_to_rel_handler.py
```diff
@@ -0,0 +1,220 @@
...
+        mapping: Mapping,
+        **_,
+    ) -> None:
+        super().__init__(combination, mapping)
+
+        self._remainder = 0
+
+        self._transform = Transformation(
+            max_=256,
+            min_=-256,
```
jonasBoss, line 114:
```suggestion
            max_=mapping.rel_input_cutoff,
            min_=-mapping.rel_input_cutoff,
```

jonasBoss: or is there a specific reason why you used +-256?

sezanzeb: Thanks.

Here is an example for when I set my mouse to the fastest speed and move it really fast:

```
Event: time 1666043095.466654, type 2 (EV_REL), code 0 (REL_X), value -243
Event: time 1666043095.466654, type 2 (EV_REL), code 1 (REL_Y), value -16
Event: time 1666043095.466654, -------------- SYN_REPORT ------------
Event: time 1666043095.474654, type 2 (EV_REL), code 0 (REL_X), value -394
Event: time 1666043095.474654, type 2 (EV_REL), code 1 (REL_Y), value -52
Event: time 1666043095.474654, -------------- SYN_REPORT ------------
Event: time 1666043095.482655, type 2 (EV_REL), code 0 (REL_X), value -378
Event: time 1666043095.482655, type 2 (EV_REL), code 1 (REL_Y), value -40
Event: time 1666043095.482655, -------------- SYN_REPORT ------------
Event: time 1666043095.490655, type 2 (EV_REL), code 0 (REL_X), value -370
Event: time 1666043095.490655, type 2 (EV_REL), code 1 (REL_Y), value -31
Event: time 1666043095.490655, -------------- SYN_REPORT ------------
Event: time 1666043095.498654, type 2 (EV_REL), code 0 (REL_X), value -356
Event: time 1666043095.498654, type 2 (EV_REL), code 1 (REL_Y), value -27
Event: time 1666043095.498654, -------------- SYN_REPORT ------------
Event: time 1666043095.506654, type 2 (EV_REL), code 0 (REL_X), value -326
Event: time 1666043095.506654, type 2 (EV_REL), code 1 (REL_Y), value -20
Event: time 1666043095.506654, -------------- SYN_REPORT ------------
Event: time 1666043095.514654, type 2 (EV_REL), code 0 (REL_X), value -354
Event: time 1666043095.514654, type 2 (EV_REL), code 1 (REL_Y), value -25
```

It's not hard at all to produce values higher than 100.

I was about to suggest to increase `rel_input_cutoff`, but it seems it is not really a cutoff. Decreasing `rel_input_cutoff` would make the values from Transformation exceed the value of 1, if the input value is high enough, right? So it is more like an inverse-speed factor

jonasBoss: No, it is not really a cutoff `Transformation` will happily create values beyond +-1.

For the `RelToRelHandler` It just scales the transformation to a reasonable input range, such that the deadzone and the non-linearity are not to small or to wide.

But more importantly for the `RelToAbsHandler` it is a cutoff and it ensures we can reach +-100% on the ABS output of the mapping. I chose 100 as the default because for me and my mouse that seems to be reasonable speed.

I think we need to chose the default cutoff such that the `RelToAbsHandler` has a good sensitivity for +-100% output. And just make sure the deadzone setting for `RelToRelHandler` feels ok.

jonasBoss:
Why do we need to differentiate between `rel_xy...` and `rel_wheel...`?

sezanzeb:
- because wheels are much faster than mice with small values. If there is only one default value, then either wheel injection will be too fast, or mouse movement too slow, depending on the default.
- Macros might contain wheel and mouse macros, by using two different values they can be controlled independently

jonasBoss:
yea, I guess this is the easiest solution for different default values.

inputremapper/injection/mapping_handlers/rel_to_rel_handler.py
```diff
@@ -0,0 +1,188 @@
...
+    def reset(self) -> None:
+        pass
+
+    def _write(self, value: float) -> None:
+        """Inject."""
+        # value is between 0 and 1, scale up
+        if self.mapping.is_wheel_output():
+            scaled = value * self.mapping.rel_wheel_speed
+        elif self.mapping.is_high_res_wheel_output():
+            scaled = value * self.mapping.rel_hi_res_wheel_speed
```
jonasBoss, line 152: There was this https://github.com/sezanzeb/input-remapper/issues/354#issuecomment-1079728380 issue where wayland can not handle HI_RES wheel events. So in the AbsToRelHandler I made sure to always send both events in the correct ratio 1/120.

If the mapping output is `WHEEL_HI_RES` the I would send the `WHEEL_HI_RES` event an accumulate output-value/120 until it is >= 1 and then also send `WHEEL` event.
If the mapping output is `WHEEL` I would always send the `WHEEL` event and 120*value as `WHEEL_HI_RES`.

jonasBoss:
> * because wheels are much faster than mice with small values. If there is only one default value, then either wheel injection will be too fast, or mouse movement too slow, depending on the default.
> 
> * Macros might contain wheel and mouse functions at the same time, by using two different values they can be controlled independently

I am a bit concerned how we can properly communicate all the different values in the GUI. Did you try using scaling values?

sezanzeb:
I thought the values in the mapping are just for advanced tweaking, since the gain controls the speed

inputremapper/injection/mapping_handlers/rel_to_rel_handler.py
sezanzeb: thanks, I implemented it

sezanzeb:
this is ready as far as I am concerned

jonasBoss:
There is something really wrong. I just created this mapping:
```json
{
    "2,1,0": {
        "target_uinput": "mouse",
        "output_type": 2,
        "output_code": 1,
        "mapping_type": "analog",
        "deadzone": 0.0,
        "gain": 0.5
    },
    "2,0,0": {
        "target_uinput": "mouse",
        "output_type": 2,
        "output_code": 0,
        "mapping_type": "analog",
        "deadzone": 0.0,
        "gain": 0.5
    }
}
```
and now my mouse is all over the place. sometimes fast, sometimes slow seems like python is to slow to compute everything in time.

inputremapper/injection/mapping_handlers/rel_to_rel_handler.py
```diff
@@ -0,0 +1,220 @@
...
+#
+# input-remapper is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with input-remapper.  If not, see <https://www.gnu.org/licenses/>.
+
+import asyncio
```
jonasBoss, line 21:
```suggestion
import asyncio
import math
```

inputremapper/injection/mapping_handlers/rel_to_rel_handler.py
```diff
@@ -0,0 +1,220 @@
...
+        scaled = value * self._scale
+        floored = int(scaled)
+        self._remainder += scaled - floored
+        if abs(self._remainder) >= 1:
+            output_value = int(scaled + self._remainder)
+            self._remainder = self._remainder - output_value
+        else:
+            output_value = floored
+
+        return output_value
```
jonasBoss, line 78:
```suggestion
        scaled = value * self._scale + self._remainder
        self._remainder = math.fmod(scaled, 1)
        return int(scaled)
```
This seems to solve the performance issues with the re-2-rel handler.

jonasBoss (changes requested):
Didn't look at everything yet, but the remainder calculation needs to be changed.

inputremapper/injection/mapping_handlers/rel_to_rel_handler.py
sezanzeb: It's a 19.5% performance increase for me

sezanzeb:
I increased the mapping rates, because those are the rates I observe on my mouse when moving it.

jonasBoss:
can you have a look at #516? I think it is better to have fewer mapping parameters especially because the different variations for the wheel essentially do the same thing.

sezanzeb:
it's unknown how fast input devices produce events without doing some observations during runtime, which makes devices behave weird when injections start.

Therefore rel2rel uses fractional input-speed calculation now. rel_speed won't have any effect.

See the docstring of `def notify` for more information

sezanzeb:
since `rel_rate` is used in rel_to_btn, this problem might affect that as well

jonasBoss:
Here is a diff which fixes typing for `InjectorMessage` and `InjectorState`
[typing.diff.txt](https://github.com/sezanzeb/input-remapper/files/9887638/typing.diff.txt)

sezanzeb:
I'll fix the tests later.

I feel comfortable with the current solution for rel2rel, rel2abs and abs2rel included in this pr

sezanzeb:
I'd like to merge this. Any changes to the configuration of speed and such please in a separate merge request

sezanzeb:
After getting https://github.com/sezanzeb/input-remapper/pull/497, https://github.com/sezanzeb/input-remapper/pull/505 and https://github.com/sezanzeb/input-remapper/pull/519 merged, I'll take a break.

I want to make the 2.0 release in January or something, which also includes https://github.com/sezanzeb/input-remapper/issues/495 and https://github.com/sezanzeb/input-remapper/issues/509.

inputremapper/injection/mapping_handlers/rel_to_rel_handler.py
```diff
@@ -0,0 +1,281 @@
...
+def is_wheel(event) -> bool:
+    return event.type == EV_REL and event.code in (REL_WHEEL, REL_HWHEEL)
+
+
+def is_high_res_wheel(event) -> bool:
+    return event.type == EV_REL and event.code in (REL_WHEEL_HI_RES, REL_HWHEEL_HI_RES)
+
+
+class Remainder:
+    _scale: int
```
jonasBoss, line 62:
```suggestion
    _scale: float
```
according to the __init__  function

inputremapper/injection/mapping_handlers/rel_to_rel_handler.py
```diff
@@ -0,0 +1,281 @@
...
+
+    _input_event: InputEvent  # the relative movement we map
+
+    _max_observed_input: float
+
+    _transform: Transformation
+
+    # infinite loop which centers the output when input stops
+    _recenter_loop: Optional[asyncio.Task]
+    _moving: asyncio.Event  # event to notify the _recenter_loop
```
jonasBoss, line 90:
```suggestion
```
not used

jonasBoss (approved):

inputremapper/injection/mapping_handlers/rel_to_rel_handler.py
sezanzeb: thanks

PR #508, opened 2022-10-22
Fix #507

### LINKED ISSUE #507: Device names containing `/` cause problems
> Here's an icky situation I am facing wrt presets: IR identifies my mouse as "Logitech M585/590." Yep, you've read that right: there's a slash in there, and that confuses the hell out of my system. As a result, quite often IR doesn't load the proper profile upon reboot, so I have to load it manually. Is there any way we can control the name of the devices as far as IR is concerned?

_Originally posted by @Aqua1ung in https://github.com/sezanzeb/input-remapper/discussions/504_

jonasBoss:
I would hold back on merging this, until we can confirm it actually fixes #507

### DIFF
--- a/inputremapper/configs/paths.py
+++ b/inputremapper/configs/paths.py
@@ -86,12 +86,25 @@ def remove(path):
         os.remove(path)
 
 
+def sanitize_path_component(group_name: str) -> str:
+    """replace characters listed in
+    https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words
+    with an underscore
+    """
+    for c in '/\\?%*:|"<>':
+        if c in group_name:
+            group_name = group_name.replace(c, "_")
+    return group_name
+
+
 def get_preset_path(group_name=None, preset=None):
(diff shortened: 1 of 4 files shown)


sezanzeb:
Thanks. What about preset names that contain "/"?

jonasBoss:
I guess we want to check those as well. But at a different place. 

The user can't do anything about a device with illegal characters. This fix will just silently change the folder name, but still show the original device name in the gui.

I don't think we want that for the preset name, because the user has control over it. There it would be better to just not create such a preset and warn the user about the illegal characters. Or maybe we can directly control the name input field to disallow illegal characters.

I think this is something we should do on the beta branch, I don't see the need to do it on main.

sezanzeb:
lets just merge it, wait and see

sezanzeb:
please merge it into beta

sezanzeb:
> the merge of main into beta is somewhat problematic with this, because the tests need to be different

I'm working on it

Issue #501, opened 2022-10-18
If there is no input to the UI for a long time, stop the helper


sezanzeb:
To avoid exposing user-input on keyboards if the GUI is left open in the background

Another call to pkexec would be needed to start a new helper

PR #519, opened 2022-10-28
Making ReaderService terminate automatically after some time

sezanzeb:
closes: https://github.com/sezanzeb/input-remapper/issues/501 and #527

- "Helper" is renamed to "ReaderService", "Reader" is renamed to "ReaderClient"
- after a timeout, the ReaderService will terminate
- the ReaderClient will restart it when needed, opening a new pkexec window
- the message-broker send method is now called "publish", which is also the term used in other message brokers (mqtt, kafka, possibly others)
- fixed: migrations were modifying the modification-date of presets, causing tests to start with different presets selected
- fixes high cpu usage when unplugging a device during reading


sezanzeb:
> Maybe we can change it such that the reader-service does not start at application startup.
> 
> If I just want to apply a different preset or edit a macro I don't need the reader-service. This way there is no password prompt on startup.

yeah, that's an easy change. I just pushed it, please try it out

jonasBoss:
well, maybe it is not the best idea: not all devices show up because the reader-service is also used to read which devices are connected, and some require root permissions.

I think I am in favor of starting the reader-service at startup, otherwise people think input-remapper does not detect all devices.

sezanzeb:
> well, maybe it is not the best idea: not all devices show up because the reader-service is also used to read which devices are connected, and some require root permissions.
> 
> I think I am in favor of starting the reader-service at startup, otherwise people think input-remapper does not detect all devices.

oh, right. Somehow on my computer this works without root permissions, I might have screwed up my permissions at some point

jonasBoss (approved):
I think we should revert the last commit and launch the reader-service at startup. Otherwise this looks good to me.

EDIT: I guess you where faster than me :)

sezanzeb:
Thank yu

PR #524, opened 2022-10-31
Prevent the daemon from choking with unsanitary device names

tsantzi:
A tiny fix that replaces path-unfriendly characters in device names such as my Logitech's "M585/590 Mouse". The slash in the name caused the daemon to fail when loading presets.

Not sure if this is the best way to fix this, I'm very new to input-remapper (only installed it just now). BTW it looks great and exactly what I was looking for, thanks for this!

### DIFF
--- a/inputremapper/daemon.py
+++ b/inputremapper/daemon.py
@@ -43,7 +43,7 @@ from inputremapper.configs.preset import Preset
 from inputremapper.configs.global_config import global_config
 from inputremapper.configs.system_mapping import system_mapping
 from inputremapper.groups import groups
-from inputremapper.configs.paths import get_config_path, USER
+from inputremapper.configs.paths import get_config_path, sanitize_path_component, USER
 from inputremapper.injection.macros.macro import macro_variables
 from inputremapper.injection.global_uinputs import global_uinputs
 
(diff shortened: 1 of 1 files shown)


inputremapper/daemon.py
```diff
@@ -437,7 +437,7 @@ def start_injecting(self, group_key, preset):
         preset_path = os.path.join(
             self.config_dir,
             "presets",
-            group.name,
+            sanitize_path_component(group.name),
             f"{preset}.json",
         )
```
jonasBoss, line 442: can you test if 
```python
        preset_path = get_preset_path(group.name, preset)
```
works here instead? 
`get_preset_path` is in `inputremapper.configs.paths`

tsantzi: This way it does the sanitization correctly, but now it fails to load the profile because it tries to locate it in the root user's home dir, so no joy.

It looks as if the unpatched/unsanitized way uses `daemon.config_dir`, which correctly reports the non-root user, while `get_preset_path` uses `CONFIG_PATH`, which reports `root` (via `pwd.getpwnam(user).pw_dir`). Maybe pass the config dir as an arg to `get_preset_path` so both daemon and client(s) can use it? I don't want to break anything, happy to look into it but it may take a while to study the code further.

jonasBoss: You are right, the service is running as root. Which is why we didn't use `get_preset_path`. I think your solution is fine as is.

jonasBoss:
@sezanzeb apparently I am not allowed to merge or push this into main. Which I missed during the merge.

So this needs to be cherry-picked.

PR #525, opened 2022-11-01
Sanitize preset name

sezanzeb:


sezanzeb:
It will just sanitize on save automatically, without doing something fancy like *checking the input and then alerting if illegal characters are used", which is the easiest solution

PR #536, opened 2022-11-12
Pylint, typehints

sezanzeb:


.pylintrc
```diff
@@ -15,3 +15,6 @@ disable=
 
     # https://github.com/psf/black/blob/main/docs/compatible_configs/pylint/pylintrc
     C0330, C0326
+
+    # pydantic uses cls as first argument
+    no-self-argument
```
jonasBoss, line 20: did you try if this works instead? https://pypi.org/project/pylint-pydantic/

sezanzeb: I haven't, thanks for pointing out

PR #549, opened 2022-11-19
Showing the input during GUI usage

### LINKED ISSUE #546: Recording output
for "Key or Macro" output add a button to record output. Should work the same way as the input-recording, but instead write the `KEY_` constants into the output text

because it is absolutely user-unfriendly to figure out how to inject `` ` `` on german keyboards

sezanzeb:
closes #546

- Removed files accidentally merged in from main https://github.com/sezanzeb/input-remapper/commit/eff463e6279d54db3e4776bcbe105fd094210299
- tests can stop all debounced functions, or run them without waiting

image

The text on the bottom can be selected and copied. This is insanely useful


inputremapper/gui/components/editor.py
```diff
@@ -367,9 +477,14 @@ def _toggle_line_numbers(self):
...
-        self._controller.save()
+        # This helps to keep the gui data up-to-date when changed-events are
+        # debounced
+        self._controller.update_mapping(output_symbol=self.code)
 
+    @debounce(1000)
     def _on_gtk_changed(self, *_):
+        # This triggers for each typed character, will cause disk-writes and writes
+        # tons of logs, so this is debounced a bit
         self._controller.update_mapping(output_symbol=self.code)
```
jonasBoss, line 488: I don't think it is a good idea do delay the mapping update. This will also delay all updates to other Gui components because they are not notified by the MessageBroker.

I know this has probably no impact at the moment because there are no other gui components which care about the `output_symbol`. But we might create some in the future. 

What happens if the preset is changed before the function is called?

The controller is currently responsible for loading and saving presets. We can create a debounced save() method in the controller and use that to avoid excessive disk-writes. 
The controller also knows when a new group or preset is loaded, so it can cancel the debounce and save the preset before loading a new one.

sezanzeb: every single typed character writes

18:51:10.085774 44307 GUI DEBUG autocompletion.py:102: get_incomplete_parameter text: aa match: <re.Match object; span=(0, 2), match='aa'>
18:51:10.984839 44307 GUI DEBUG message_broker.py:71: from data_manager.py:128: Signal=mapping: {'event_combination': <EventCombination (1, 276, 1)>, 'target_uinput': 'keyboard', 'output_symbol': 'aa', 'mapping_type': 'key_macro'}
18:51:10.987010 44307 GUI DEBUG message_broker.py:71: from controller.py:619: Signal=status_msg: StatusData(ctx_id=5, msg='Mapping error at Button EXTRA, hover for info', tooltip='3 validation errors for Mapping\noutput_symbol\n  The output_symbol "aa" is not a macro and not a valid keycode-name (type=value_error)\n__root__\n  Missing Argument: Mapping must either contain `output_symbol` or `output_type` and `output_code` (type=value_error)\n__root__\n  missing macro or key: "Button EXTRA" is not used as analog input, but no output macro or key is programmed (type=value_error)')
18:51:10.987373 44307 GUI INFO preset.py:184: Saving preset to /home/mango/.config/input-remapper/beta_1.6.0-beta/presets/USB Optical Mouse/a.json

to the logs

when leaving the text input by interacting with any other gui component in any way, saving is tirggered immediately via `_on_gtk_focus_out`.

> What happens if the preset is changed before the function is called?

Most of the time nothing. I don't like both options now though

jonasBoss: yeah the MessageBroker is quite verbose debug_logging every single message. But using it less is the wrong "solution". Creating more debug levels seems more appropriate.

sezanzeb: I pushed a change that prevents stuff from happening when the mapping changes (caused by a preset change), with tests

sezanzeb: I'll make a separate PR and merge this without the debounce

PR #550, opened 2022-11-20
Input event with origin

### LINKED ISSUE #435: undefined behaviour if multiple devices in the same group have similar capabilities (beta)
The Sony DualSense controller has a mouse-pad which reports `ABS_X` and `ABS_Y` as well as the standard controller layout which also reports `ABS_X` and `ABS_Y` events.
When a mapping uses one of those events as input both the joystick and the mouse-pad can trigger this mapping.

This leads to all sorts of issues, especially because the `absinfo` are not the same for those axis.

jonasBoss:
fix #435

- [x] update InputEvent to include origin and analog_threshold
- [x] save preset as list with InputEvent as dict inside the mapping
- [x] add capability to load old beta preset
- [x] add origin info to each InputEvent in the EventReader
- [x] update EventHandler's
- [ ] ~~make sure forward_release forwards to the correct device~~ #577 
- [x] update event pipline creation to respect origin information
- [x] add fallback when origin is not set in mapping
- [x] grab devices based off of origin
  - [x] add fallback when origin can not be found during device grabbing???
- [x] seperate InputEvent from InputConfig
- [x] tests
- [x] delete old stuff
- [x] Update documentation (the preset json has changed)


jonasBoss:
So this is actually more than just including a origin with each event.

The main change is actually the Separation of InputEvent and InputConfig. Currently they are both very similar but it gives us much more flexibility as the InputConfig is no longer restricted by a single event or performance concerns. This can help if we ever choose to process the input on a per frame basis instead of a per event basis. 
We can use the InputConfig to store mapping parameters which can be individual for each input of a mapping (e.g. the `release_combination_keys` flag).

I also renamed the EventCombination to InputCombination.
@sezanzeb I would like to hear your thoughts on this change.

sezanzeb:
I'll look at it soon

sezanzeb:
A miroboard showing how all of our models and classes are related might be great at this point

sezanzeb:
ah,

so InputConfig is for the mapping and describes the input for a mapping

and InputEvent is for the EventHandler pipeline and describes an actual event that is happening right now

jonasBoss:
yeah, that is the idea. 
Pylint comes with a `pyreverse` command which can be used to generate UML. Although, Pylint seems to be broken for me :(

sezanzeb:
graph

sezanzeb:
How about renaming `InputCombination` to `InputConfigCombination`?

inputremapper/input_event.py
```diff
@@ -62,103 +47,62 @@ class InputEvent:
...
-            except InputEventCreationError:
-                pass
+        raise TypeError(f"cannot compare {type(other)} with InputEvent")
 
-        if event:
-            return event
-
-        raise ValueError(f"failed to create InputEvent with {init_arg = }")
+    @property
+    def input_match_hash(self) -> Hashable:
```
sezanzeb, line 67: How about adding a `__hash__` function to both InputConfig and InputEvent?

Instead of using `input_match_hash` I would probably prefer a function like

```py
def is_matching_event(event: Inputevent): bool
```

in InputConfig or something.

sezanzeb: So that

```py
context.notify_callbacks[input_config].append(
    handler.notify
)
```

instead of 

```py
context.notify_callbacks[input_config.input_match_hash].append(
    handler.notify
)
```

sezanzeb:
```py
context.add_notify_callback(input_config, handler.notify)
```

instead of

```py
context.notify_callbacks[input_config.input_match_hash].append(
    handler.notify
)
```

sezanzeb:
```py
notify_callbacks = self.context.notify_callbacks.get(event.input_match_hash)
```

instead of

```py
notify_callbacks = self.context.get_notify_callbacks(event)
```

inputremapper/configs/input_config.py
```diff
@@ -0,0 +1,347 @@
...
+    ecodes.KEY_RIGHTSHIFT,
+    ecodes.KEY_LEFTCTRL,
+    ecodes.KEY_RIGHTCTRL,
+    ecodes.KEY_LEFTALT,
+    ecodes.KEY_RIGHTALT,
+]
+
+
+class InputConfig(BaseModel):
+    """The configuration of a single input to a mapping"""
```
sezanzeb, line 46:
```suggestion
    """Describes a single input within a combination, to configure mappings."""
```

sezanzeb: Suggestion for the docstring of InputEvent:

```py
class InputEvent:
    """Events that are generated during runtime.

    Is a drop-in replacement for evdev.InputEvent
    """
```

inputremapper/injection/injector.py
```diff
@@ -212,25 +247,20 @@ def _grab_devices(self) -> GroupSources:
...
+                    if input_config.origin is None:
+                        update_configs[input_config] = input_config.modify(
+                            origin=get_device_hash(device)
+                        )
+
+            if update_configs:
+                combination = list(mapping.input_combination)
+                for old, new in update_configs.items():
+                    combination[combination.index(old)] = new
+                mapping.input_combination = combination
```
sezanzeb, line 263: Some comments would be nice here. Maybe also split it into some private methods to make it easier to read

inputremapper/gui/reader_service.py
```diff
@@ -267,68 +268,85 @@ def _create_event_pipeline(self, sources: List[evdev.InputDevice]) -> ContextDum
...
         # create a context for each source
         for device in sources:
+            device_hash = get_device_hash(device)
             capabilities = device.capabilities(absinfo=False)
 
             for ev_code in capabilities.get(EV_KEY) or ():
-                context.notify_callbacks[(EV_KEY, ev_code)].append(
+                input_config = InputConfig(
+                    type=EV_KEY, code=ev_code, origin=device_hash
+                )
```
sezanzeb, line 277: I think the constructor should hash the device instead

```py
input_config = InputConfig(
    type=EV_KEY, code=ev_code, origin=device
)
```

inputremapper/configs/preset.py
```diff
@@ -286,25 +283,44 @@ def _get_mappings_from_disc(self) -> Dict[EventCombination, MappingModel]:
...
+                        for k, v in zip(
+                            ("type", "code", "analog_threshold"), s.split(",")
+                        )
+                    }
+                    assert len(cfg) == 3
+                    config.append(cfg)
+                return config
+
+            for c, d in preset_list.items():
+                d["input_combination"] = str_to_cfg(c)
```
sezanzeb, line 308: no abbreviated variable names please

inputremapper/configs/migrations.py
```diff
@@ -258,24 +270,41 @@ def _otherwise_to_else():
...
+    for event_str in combination_string.split("+"):
+        cfg.append(
+            {
+                k: int(v)
+                for k, v in zip(
+                    ("type", "code", "analog_threshold"), event_str.split(",")
+                )
+            }
+        )
+    return InputCombination(cfg)
```
sezanzeb, line 295: no abbreviated variable names please.

Would this be an easier to read equivalent?

```suggestion
    configs = []
    for event_str in combination_string.split("+"):
        type, code, analog_threshold = event_str.split(",")
        configs.append({
            type: type,
            code: code,
            analog_threshold: analog_threshold
        })
        
    return InputCombination(configs)
```

sezanzeb (commented):
Thanks. I think splitting it into InputConfig and InputEvent makes sense

inputremapper/input_event.py
jonasBoss: The hashing is kind of difficult:
The InputConfig needs two different hashes.
 1. It is itself a hash, via the `__hash__` method. 
 2. It has the input_match_hash

The first one is needed for the Preset class, as it is totally okay to have multiple mappings with the same input, as long as they have different `analog_threshold` values. This is currently implemented with the `__hash__` method because than it is automatically used whenever we use a `InputCombination` as a hash.

The second one is needed in the event pipeline to match against incoming InputEvent's. There we must ignore the `analog_threshold`.

The InputEvent only has the input_match_hash, and is never used as a hash itself.

I would rather remove the `__hash__` method and implement one for the InputCombination instead (maybe this one can elegantly solve the Permutation check). This way InputConfig can not be a hash itself. _explicite is better than implicit_ especially in this case, where confusion is likely.


> ```python
> context.add_notify_callback(input_config, handler.notify)
> ```

This method would do the same thing as we already do:
```python
def add_notify_callback(self, input_config, callback):
    self.notify_callbacks[input_config.input_match_hash].append(callback)
```

inputremapper/gui/reader_service.py
jonasBoss: Not possible. We need to construct the InputConfig from the json file. 

Also it would make evdev.InputDevice a dependency for InputConfig, which is not really necessary.

inputremapper/input_event.py
sezanzeb: > This method would do the same thing as we already do:

What do you mean?

inputremapper/configs/migrations.py
jonasBoss: the code in preset.py is not needed. It is a migration from beta to beta. which should be removed prior to the actual release.

inputremapper/gui/reader_service.py
sezanzeb: Ok, should `origin` maybe be renamed to `origin_hash` then?

In any case, there should be a comment above the member variable explaining that this is a hash

sezanzeb: Maybe use a type alias like "DeviceHash = str", and then change

```py
def get_device_hash(device: evdev.InputDevice) -> str:
```

to

```py
def get_device_hash(device: evdev.InputDevice) -> DeviceHash
```

and

```py
origin: Optional[str] = None
```

to 

```py
origin: Optional[DeviceHash] = None
```

inputremapper/input_event.py
jonasBoss: I was just thinking, there is not much the `add_` and `get_notify_callback` methods would not do much at all. But maybe it makes still sense if we want to make changes to the Context in the future.

sezanzeb: It makes it easier to read and use, because it is not required to remember that `input_match_hash` has to be used, and no knowledge about the datastructure of the notify callbacks would be required anymore

inputremapper/injection/injector.py
jonasBoss: I just shuffled a bunch of functions around in the Injector. Hopefully it is now clearer what is going on.

sezanzeb:
Thanks. Is this ready to merge?

I would probably still rename `origin` to `origin_hash` though. What are your thoughts on that?

jonasBoss:
I will rename `origin` and then it is ready to merge.
I am not sure about the `InputCombination` it is already in the `configs.input_config` module. So it should be clear what it does. 

As for this:
> make sure forward_release forwards to the correct device

That is a bug which is present on beta since ever. I just never thought about it until this PR made it kind of obvious. 
There are two possible solutions, which both will need some work.

Issue #577, opened 2022-12-14
`release_combination_keys` does not always forward to the correct device (beta)


jonasBoss:
When a input combination has inputs from different `/dev/input/eventXX` devices the CombinationHandler will not correctly release the combination keys on the forwarded device.

sezanzeb:
I'm looking into it

Issue #547, opened 2022-11-18
Make an autoclicker with start and stop


supertriodo:
Hey there! First thanks for all the hard work in such an amazing tool. I use it everyday for all my games.

I can make an autoclicker of left click linking it to right click as long as I press it.

hold(key(BTN_LEFT).wait(200))

But, is there a way to make it so when I right click it starts the autoclick (on left click) until I press it again (right click)?

Thanks!

sezanzeb:
This requres a new macro, maybe something like `repeat_while`, which works similar to `if_eq`, but repeats the macro similar to `repeat`

sezanzeb:
Writing macros is actually fairly easy. See https://github.com/sezanzeb/input-remapper/blob/beta/inputremapper/injection/macros/macro.py#L604 and https://github.com/sezanzeb/input-remapper/blob/beta/inputremapper/injection/macros/macro.py#L447

sezanzeb:
it could then be used like

```c#
set(toggle, 1).repeat_while(
    $toggle,
    1,
    hold(key(BTN_LEFT).wait(200))
)
```

the button that stops the loop:

```c#
set(toggle, 0)
```

supertriodo:
That sounds great, I'll watch closely for when it's implemented, thanks for the answer!

Thorinori:
Bumping this, it would be a fantastic feature to have included soon, and would functionally allow input-remapper to be a reasonable replacement for AutoHotKey that works with Linux for the people that want to make clickers etc. May look into implementing it myself as well potentially.

jonasBoss:
Just leaving this here as an idea how to solve this with mapping handlers (from https://github.com/sezanzeb/input-remapper/pull/578#issuecomment-1374673063):

> Add a InputConfig.toggle parameter which can solve https://github.com/sezanzeb/input-remapper/issues/547 and https://github.com/sezanzeb/input-remapper/issues/378 by inserting a ToggleHandler in the event-pipeline

RickAndTired:
Would something like this work? (This exact example doesn't work)

Set `a` to `hold(key(BTN_LEFT).wait(200))`
Set `b` to `if_eq($toggle, 0, key_down(a).set(toggle, 1), key_up(a).set(toggle, 0))`

inakilbss:
[...]

Issue #378, opened 2022-04-21
Repeat macro until key pressed again


sezanzeb:
see https://github.com/sezanzeb/input-remapper/discussions/367

sezanzeb:
1. I think currently on main if a macro is already running, it won't trigger it again. Therefore it is not possible to interact with a running macro via a variable or something to make it stop repeating.
2. There is no `while` macro function

jonasBoss:
should be possible with a mapping handler. We can have a mapping handler which triggers the mapping when the input combination triggers once, and releases the mapping when the input combination is triggered a again. That way not only macros can benefit from this option, but also regular key or combination to key mappings. We can even use it to enable and disable the mapping of axis to another axis.

muikku:
Upvote for this feature!

Plyply99:
Upvote for this feature!

PR #578, opened 2022-12-15
Fix CombinationHandler releasing

sezanzeb:
notify of CombinationHandler has a forward_release method, which iterates over keys_to_release and sends all of them into a single forward_to device

notify passes a single "forward" uinput down.

some of the events in keys_to_release originate from other devices though

I'm trying to add a property to the inputEvent that has a reference to the correct forwarding device


jonasBoss:
I think it is a better solution to add all uinputs and sources to the Context. this way we can remove the Source and forward parameters from the notify method. yesterday did some of the changes to the context already: https://github.com/jonasBoss/key-mapper/tree/fix_release_combination_keys

sezanzeb:
To then get the correct device for forwarding via `origin_hash`?

jonasBoss:
each InputEvent and InputConfig already has a `origin_hash` so we can use that to in the `forward_release` method something like this:

``` python
    def forward_release(self) -> None:
        """Forward a button release for all keys if this is a combination

        this might cause duplicate key-up events but those are ignored by evdev anyway
        """
        if len(self._pressed_keys) == 1 or not self.mapping.release_combination_keys:
            return

        for input_config in self.mapping.input_combination:
            device = self.context.get_uinput(input_config.origin_hash)
            device.write(*input_config.type_and_code, 0)
        forward.syn()
```

jonasBoss:
This also allows us to simplify all the `AbsTo....notify()` methods, because we can retrieve the absinfo during the `__init__`

sezanzeb:
> ```python
> device = self.context.get_uinput(input_config.origin_hash)
> ```

I think get_uinput gets the target uinput though

sezanzeb:
ah, get_uinput does two things. I'll try to split it

sezanzeb:
I don't have a device that triggers that edge case, can you please try it?

jonasBoss:
> I don't have a device that triggers that edge case, can you please try it?

I don't either

jonasBoss:
> ah, get_uinput kinda does two things. I'll try to split it

Treating the `global_uinputs` and the forward_uinputs the same opens up the possibility to have mappings where the target uiput is the forwarded device. We just need to change the MappingHandler's to not use the global_uinputs, but instead the `context.get_uinput()`

inputremapper/configs/input_config.py
```diff
@@ -55,7 +55,7 @@ class InputConfig(BaseModel):
     # origin_hash is a hash to identify a specific /dev/input/eventXX device.
     # This solves a number of bugs when multiple devices have overlapping capabilities.
     # see utils.get_device_hash for the exact hashing function
-    origin_hash: Optional[DeviceHash] = None  # type: ignore
+    origin_hash: Optional[DeviceHash] = None
```
jonasBoss, line 58: I would like to ensure the origin_hash is always lowercase. It is a big hex number, so it should be case insensitive.

jonasBoss: Okay I see just now, you added a valitator for that.

inputremapper/injection/event_reader.py
```diff
@@ -60,7 +60,7 @@ def __init__(
         self,
         context: Context,
         source: evdev.InputDevice,
-        forward_to: evdev.UInput,
+        forward_to: evdev.UInput,  # TODO I think this can go as well
```
jonasBoss, line 63: And also the source. And provide the correct origin_hash instead.

sezanzeb: Thanks.

About the source: I'd generally prefer passing the original objects whenever possible, instead of derived information or members.

jonasBoss:
```python
hash_a = "some hash"

context.get_source(hash_a)  # returns the source device
context.get_uinput(hash_a)  # returns the forward_to device
```
Since we need the origin hash to retrieve the correct forward_to device we can reduce the number of arguments to `notify()` by providing only the hash.

This also eliminates the possibility to create a bug where the `source` has a different `origin_hash` than the one provided. 

We could instead not provide the origin_hash but calculate it during runtime, but that is expensive.

tests/lib/fixtures.py
```diff
@@ -311,13 +311,12 @@ def QuxSlashDeviceQuestionmark(self):
 def get_combination_config(
     *event_tuples: Tuple[int, int] | Tuple[int, int, int]
 ) -> Iterable[Dict[str, int]]:
-    """convenient function to get a iterable of dicts, InputEvent.event_tuple's"""
-
+    """Convenient function to get an iterable of dicts, InputEvent.event_tuples."""
     for event in event_tuples:
         if len(event) == 3:
-            yield {k: v for k, v in zip(("type", "code", "analog_threshold"), event)}
+            yield {"type": event[0], "code": event[1], "analog_threshold": event[2]}
```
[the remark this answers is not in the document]
sezanzeb: lies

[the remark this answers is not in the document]
jonasBoss: indeed, looks like mypy does not pick up on the type guard. 
maybe you can do something like this:
```python
config = {"type": event[0], "code": event[1]}
try:
    config["analog_threshold"] = event[3]
except IndexError:  # or whatever the error is called
    pass
yield config
```

sezanzeb:
I'd only use tuples for things that have a fixed number of elements.

For combinations and such, I would rather use a list of InputConfigs, since the number of InputConfigs varies.

The `InputCombination` constructor now always expects a list of InputConfigs. A call like `InputCombination(InputConfig)` will fail. I find this more clear. This preference has also to do with my past experience on a different project. Functions/constructors that are flexible with the argument type were often a cause of major headaches.

I also realized that I prefer classmethods over factories.

sezanzeb:
Do you find

```py
InputConfig(type=EV_ABS, code=ABS_X)
```

or

```py
InputConfig.abs(ABS_X)
```

more beautiful?

The classmethod
- clearly defines which arguments are needed for this type of event (analog_threshold is only required for abs)
- can contain a custom docstring that matches the event type
- is shorter with less special characters
- doesn't require importing EV_ABS

inputremapper/configs/input_config.py
```diff
@@ -299,7 +377,17 @@ def empty_combination(cls) -> InputCombination:
 
         Useful for the UI to indicate that this combination is not set
         """
-        return cls([{"type": 99, "code": 99, "analog_threshold": 99}])
+        return cls([{"type": EMPTY_TYPE, "code": 99, "analog_threshold": 99}])
```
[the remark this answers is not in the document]
sezanzeb: `List[Dict[str, int]]` seems to be pretty much compatible to `Union[Iterable[Dict[str, Union[str, int]]], Iterable[InputConfig]]`. Am I missing something?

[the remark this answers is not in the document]
jonasBoss: I think it is compatible, don't see a reason why it should not be.

inputremapper/configs/input_config.py
```diff
@@ -299,7 +377,17 @@ def empty_combination(cls) -> InputCombination:
...
+
+    @classmethod
+    def from_tuples(cls, *tuples):
+        """Construct an InputCombination from (type, code, value) tuples."""
```
jonasBoss, line 384:
```suggestion
        """Construct an InputCombination from (type, code, analog_threshold) tuples."""
```

jonasBoss:
> I'd only use tuples for things that have a fixed number of elements, like coordinates.
>
> So for combinations, I would rather use a list of InputConfigs instead of a tuple, since the number of InputConfigs in there can be anything.

The `InputCombination` is used as a hash in the `Preset` class, so it needs to be immutable, which is why it is a tuple. Sure, there can be combinations of different lengths, but a given combination can not change it's length.

> The `InputCombination` constructor now always expects a list of `InputConfigs`. A call like `InputCombination(InputConfig)` will fail.

Yeah, that makes more sense. I think this was a leftover from way back when there was the `Key` class.

> I also realized that I prefer classmethods over factories.

I think a factory should always be a classmethod. 

> Since the `InputEvent` constructor requires the sec and usec parameters, which is why `new_event` was added, and I definitely prefer `InputEvent.key(1, 1)` over `new_event(EV_KEY, 1, 1)`, having `InputConfig.key(1)` just seems consistent with the rest of the code.

The `InputConfig` is a pydantic model, so we have the default init anyway, and it is very convenient to create a `InputConfig` from a dict or a json file.
I am fine with adding more thins like `InputEvent.rel(...)` for the `InputEvent` class because its just a more convinient `evdev.InputEvent`.
 
But I think we should not do this for the `InputConfig` class. The `InputConfig` will probably be expanded and changed in the future. 

Here are some Ideas how the `InputConfig` could change:
 - A obvious change would be to move the `Mapping.release_combination_keys` from the mapping to the `InputConfig`. This will allow us to specify if the release event should be forwarded for each input individually.
 - There are a few more parameters which might be a good idea to move to the `InputConfig` e.g. `Mapping.release_timeout` and `Mapping.force_release_timeout`
 - Add a `InputConfig.toggle` parameter which can solve #547 and #378 by inserting a `ToggleHandler` in the event-pipeline

sezanzeb:
> The `InputCombination` is used as a hash in the `Preset` class, so it needs to be immutable, which is why it is a tuple. Sure, there can be combinations of different lengths, but a given combination can not change it's length.

makes sense

the motivation was that in some cases

```py
InputCombination([InputConfig(type=EV_KEY, code=10)])
```

was better readable than

```py
InputCombination((InputConfig(type=EV_KEY, code=10)))
```

because the square brackets are easier to point out, or something. And it shows better that the parameter can be extended with more InputConfigs if wanted.

So this is just from a usage point of view. Internally it can stay a tuple

Do you have some more thoughts about it from that point of view?

jonasBoss:
> Do you have some more thoughts about it from that point of view?

They are both valid and will work. Which is also indicated by the Typing:

```python
InputCombinationInit = Union[
    Iterable[Dict[str, Union[str, int]]],
    Iterable[InputConfig],
]

class InputCombination(Tuple[InputConfig, ...]):
# this is equivalent to 
# class InputCombination(tuple):
#
# but also provides the type of tuple we are inheriting

    def __new__(cls, configs: InputCombinationInit) -> InputCombination:
        ...
```

You should be able to use 
```python
InputCombination([InputConfig(type=EV_KEY, code=10)])
```
wherever you like. And I agree it is easier to read.
It might be marginally slower than using a Tuple, but that does not matter since this is a config and not a event.

sezanzeb:
TODO: run integration tests locally

sezanzeb:
TODO: there are a few `TODO test` markings added that needs to be taken care of

sezanzeb:
To reiterate, because I might have forgotten what this was actually all about

The point of this is, that the input that triggers a combination is released correctly, when a device has sub-devices with similar capabilities

image

a mouse "A" that has two devices "A_1" and "A_2" in evdev. Both support BTN_EXTRA and BTN_SIDE for whatever funny reason.

- BTN_EXTRA and BTN_SIDE arrive via "A_2"
- they are sent into "A_2 forwarded"
- the mapping triggers
- the release events for those two go into device "A_2 forwarded", and not into "A_1 forwarded"

Was that it?

jonasBoss:
this closes #577

The issue is more general than what you described.
If a mapping has inputs from two different sub devices, it will always fail to send the release events to the correct one. Regardless if both sub devices have similar capabilities or not.

Currently the `notify` method looks like:
```python
def notify(
    self,
    event: InputEvent,

    # the source of the event, relevant to get the absinfo for AbsTo...Handler
    source: evdev.InputDevice,

    # the forward device which is blindly used to forward the release events
    # on mappings with multiple input devices, this is guaranteed to be the wrong one 
    # for at least one input
    forward: evdev.UInput = None, 
    suppress: bool = False,
) -> bool:
```

If we manage the forward devices in the `Context` we can use the `origin_hash` from the mapping 
to find the correct forward device, and remove the `forward` argument from `notify`.

That requires all Handlers to have access to the context.

It would also make senses store references to the source devices and the virtual uinputs in the context. Than we can also remove the source argument and we can hide the fact that there is a `global_uinputs` object.

sezanzeb:
Thanks!

CombinationHandler has the context, and it figures out the device to send the release to by looking at the origin_hash. Sounds good.

I just need to write a few tests for this and document the reasoning and then this is done. All old tests are passing.

inputremapper/injection/event_reader.py
sezanzeb: > Since we need the origin hash to retrieve the correct forward_to device we can reduce the number of arguments to notify() by providing only the hash.

the correct forward_to device is resolved by looking at the InputConfig objects though

jonasBoss (approved):

inputremapper/injection/event_reader.py
jonasBoss: That is correct, so we don't even need the hash.

It is not relevant to this PR, but I was looking through all the `notify()` methods, and the source is only ever used to do some init stuff like finding out the `absinfo`.

Now that the context knows about all the sources and it gets passed to all `MappingHandlers` we should probably remove the `source` argument form `notify` and do the init stuff inside `__init__` that would clean up a few mapping handlers.

inputremapper/configs/input_config.py
```diff
@@ -235,35 +257,62 @@ def _remove_analog_threshold_for_key_input(cls, values):
...
     def __new__(cls, configs: InputCombinationInit) -> InputCombination:
-        if isinstance(configs, InputCombination):
-            return super().__new__(cls, configs)  # type: ignore
+        """Create a new InputCombination.
+
+        Examples
+        --------
+            InputCombination([InputConfig, ...])
+            InputCombination([{type, code, value}, ...])
+        """
```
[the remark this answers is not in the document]
sezanzeb: Thanks

PR #592, opened 2023-01-10
Single-Threaded Service

jonasBoss:
This runs the Service in a single process. 
It should reduce overhead (only one python interpreter running) and it allows for better Profiling. In addition it allows us to simplify the injector, shared-dict and test-fixtures because we can remove all the communication stuff.

There is one issue with this: the use of a different dbus library [dbus-next](https://github.com/altdesktop/python-dbus-next). This is needed because the current library pydbus needs a GLib.MainLoop, which is incompatible with asyncio. There is a [merge-request](https://gitlab.gnome.org/GNOME/pygobject/-/merge_requests/189) to add a asyncio compatible GLib.MainLoop, but who knows when this is ready.

Unfortunately dbus-next is not part of the Ubuntu repos. 
An alternative could be [dbussy](https://github.com/ldo/dbussy) but that deliberately introduces inconsistency with method return types ([see this issue](https://github.com/ldo/dbussy/issues/30)). 
There is also [jeepney](https://gitlab.com/takluyver/jeepney), but that is missing a high level interface, so it will require extra work.
Both of these alternatives are part of the ubuntu repos. But I still think that dbus-next offers the best interface with only one [issue](https://github.com/altdesktop/python-dbus-next/issues/119) for which I found a workaround.

So the question is:
 - do we want this?
   - I think yes, it will make the code simpler to reason about.
 - Is a single process fast enough to handle multiple devices with high polling rates?
   - [ ] this needs testing
 - do we want to, and can we package InputRemapper with dbus-next?

### DIFF
--- a/bin/input-remapper-control
+++ b/bin/input-remapper-control
@@ -131,11 +131,11 @@ def communicate(options, daemon):
             logger.info('Autoloading all')
             # timeout is not documented, for more info see
             # https://github.com/LEW21/pydbus/blob/master/pydbus/proxy_method.py
-            daemon.autoload(timeout=10)
+            daemon.autoload()
         else:
             group = require_group()
             logger.info('Asking daemon to autoload for %s', options.device)
-            daemon.autoload_single(group.key, timeout=2)
+            daemon.autoload_single(group.key)
 
     if options.command == START:
         group = require_group()
--- a/bin/input-remapper-service
+++ b/bin/input-remapper-service
@@ -50,5 +50,4 @@ if __name__ == '__main__':
         log_info('input-remapper-service')
 
     daemon = Daemon()
-    daemon.publish()
     daemon.run()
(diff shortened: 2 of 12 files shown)


sezanzeb:
> can we package InputRemapper with dbus-next?

This is a question for @skitt :)

would this be problematic for the ubuntu package?

> Is a single process fast enough to handle multiple devices with high polling rates? 

Multithreading in python is a bit special apparently. GIL prevents the usage of more than one core. So while multithreading helps to make things concurrent when threads are waiting for i/o, it does not improve performance.

I wonder what the benefits of multithreading over asyncio are.

sezanzeb:
I think it might make sense to merge this after the 2.0 release

jonasBoss:
> I think it would make sense to release 2.0 before we merge this

I think so as well

skitt:
From a packaging perspective dbus-next will be available in Debian 12 and is available in Ubuntu 22.10. So the distro packages will be OK, but the package provided on GitHub will be usable on a limited selection of releases.

Issue #448, opened 2022-07-29
Custom config directory 


tpeacock19:
The configuration directory can be set by using `--config-dir` with `input-remapper-control`. However, this option is not offered with `input-remapper-gtk` or any of the other executables. This makes it impossible to edit the presets located in those directories.

I came across this because I have a non-standard `xdg-config-home` among other directories.  So this defaulting to `.config/input-remapper` is a bit annoying for me.

I tried looking into the codebase to see how easily it would be to use the actual `XDG_CONFIG_HOME` variable but it seems to be a bit more involved than just a simple edit to the `user.py`. I think it would be great to incorporate the actual `XDG_CONFIG_HOME` variable as well as allowing the user to set the config directory when running `input-remapper-service`.

jonasBoss:
Did you look at the beta source? I am not sure how it is set up on main, but it should be relative simple to implement this in beta:

1. I think it makes sense to merge `user.py` into `configs/paths.py` they are both short and closely related.
2. update `paths.py` because `.config/input-remapper` is currently hard-coded in line 33
3. go through the source and make sure we always use the `HOME`, `USER` and  `CONFIG_PATH` constants form `paths.py` and not some other hard-coded paths.

yngmjgsd:
If you're looking for usage of XDG_CONFIG_HOME explicitly then this edit will do this in beta and it will be using XDG_CONFIG_HOME if specified explicitly. Though, it doesn't provide any GUI option.

yngmjgsd:
Sorry, I messed up with a previous request. Now I fixed it. Again, this is just an example of what can be done. It appears to me that paths.py will be used by the daemon as well.

PR #595, opened 2023-01-12
Settings up XDG_CONFIG_HOME variable as a path for input-remapper configs explicitly if found in a user environment.

yngmjgsd:
Settings up XDG_CONFIG_HOME variable as a path for input-remapper configs explicitly if found in a user environment. A possible solution for the 'Custom config directory' #448 issue.


sezanzeb:
If you are still considering to fix the pipeline problems: The config code has changed, there is no beta anymore and everything is on main now.

Issue #537, opened 2022-11-12
Fedora packaging


Sinthoras7:
There are only instructions on how to install it on Debian and Arch based systems, not Fedora.
 Can it be installed on Fedora too? If so, how?

sezanzeb:
Try the manual installation as shown in the readme. I don't use fedora, so I can't offer support.

sezanzeb:
Ideally someone who knows how this works gets input-remapper into fedoras official repo https://docs.fedoraproject.org/en-US/package-maintainers/New_Package_Process_for_New_Contributors/

Or contributes a script that creates a .rpm package file

Tesla-Tank:
> There are only instructions on how to install it on Debian and Arch based systems, not Fedora. Can it be installed on Fedora too? If so, how?

Try using this copr, I've used it for a while now and had no issues installing or using it. 
https://copr.fedorainfracloud.org/coprs/sunwire/input-remapper/

Zeit08:
Thanks, I also was looking for input-remapper for Fedora, and I also suggest it to be in the main repo as it may be a fundamental program for some of us. Unfortunately I have no skill in compiling and maintaining it in the repo...

sezanzeb:
> First, the .py files that do not end up in /usr/bin but reside in the system's site packages directory (e.g. /usr/lib/python3.11/site-packages/) should not have a shebang line (#!/usr/bin/python3)

Thanks. I Just removed them from both main and beta

> input-remapper

that would be the intended name

> inputremapper

Because python packages can't have dashes in their name. `import input-remapper` is not valid python syntax unfortunately

> key-mapper

The old name

> input_remapper

Only seen in the code to comply with the snake_case Python convention for variable names

alxpl:
Thanks, it's been so long since I last touched python code that I had completely pushed out of my memory its naming conventions.
I will get in touch with Paweł who has set up the COPR and see if he's willing to (co-)maintain the package in Fedora.
When do you think the 1.6 release will be ready?
I am a bit swamped with work at the moment, I haven't even had the time to deal with #542. This is going to be a weekend project, but once the package is up for review, things will move along reasonably fast, there's quite a few people who "owe" me package reviews.

sezanzeb:
It will be a 2.0 release, and probably in february

alxpl:
Hello,

[...]

PR #611, opened 2023-02-07
Add AppStream metadata

alxpl:
I've created a basic AppStream MetaInfo file and I've added a directive in setup.py for it to be placed in the /usr/share/metainfo/ directory.

I tried to follow the [documentation](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic) regarding the <id> tag and I replaced with an underscore the hyphen in the name, as their use is "discouraged".

As I wrote in https://github.com/sezanzeb/input-remapper/issues/537#issuecomment-1356863846, new screenshots will have to be created at a 16:9 ratio and the current minimum recommended height is 351 px (without padding).

I've left a placeholder copyright comment (the second line).

### DIFF
--- a/data/io.github.sezanzeb.input_remapper.metainfo.xml
+++ b/data/io.github.sezanzeb.input_remapper.metainfo.xml
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright [year] [name] -->
+<component type="desktop-application">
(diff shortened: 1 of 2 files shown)


alxpl:
P.S.: Happy new year!

sezanzeb:
Thanks. I'll make sure to update the screenshots soon in beta to be 16:9

PR #615, opened 2023-02-11
Rename desktop file to allow Dash pinning

ubunatic:
To allow pinning of the app in the GNOME Shell Dash, the binary name and desktop file name must match. (Also see my [experiment](https://github.com/ubntc/lab/tree/main/gtk/minimal-gtk3-app) on this behavior).

This change has the following effects.
* The running Gtk app can be pinned in the Dash (right click -> Pin in Dash) 
* All instances are grouped under same Dash icon,
* A pinned launcher icon and the running app share the same space in the Dash
* Instance windows are grouped under the Dash icon
* The Dash item for now uses the input-remapper icon


sezanzeb:
Thanks!

PR #616, opened 2023-02-11
Add install scripts for local development

ubunatic:
While trying to do some hands on coding for the app, I had some trouble with left-over files on the system and needed a way to clean up.

For that I added a cleanup script and also a script for installing from inside a virtual env.

If you think this is not helpful, just close this PR.
If there is a better way to install and clean up, please let me know ;)

### DIFF
--- a/scripts/inject-path.sh
+++ b/scripts/inject-path.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+# Modifies (or cleans) local bin/* files to point to the local development
+# modules by injecting a corresponding `sys.path`.
+
+case "$1" in
+    inject)
+        inject_path="${2:-$PWD}"
+        echo "inject import path '$inject_path' in bin file sources"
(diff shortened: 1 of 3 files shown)


sezanzeb:
> If there is a better way to install and clean up, please let me know ;)

If installed via `sudo python3 setup.py install` there is no better way. Thanks :)

ubunatic:
> > If there is a better way to install and clean up, please let me know ;)
> 
> If installed via `sudo python3 setup.py install` there is no better way. Thanks :)

In the docs I found that I should use `pip install .`.
I will try out the other variant, I guess I could use `sudo /usr/bin/python` to install it in the system (skipping any virtual envs here to have the service installed without hacks).
If I find a way that skips the virtual env, so that sth. like `pip uninstall` works afterwards, I will adjust the scripts.

sezanzeb:
> In the docs I found that I should use pip install ..

Yeah, sometimes this doesn't work though, I think especially in fedora if remember correctly

PR #627, opened 2023-02-20
Automatically set "Use as analog" if output is analog

sezanzeb:
closes https://github.com/sezanzeb/input-remapper/issues/495

### DIFF
--- a/inputremapper/gui/controller.py
+++ b/inputremapper/gui/controller.py
@@ -138,7 +138,8 @@ class Controller:
             self.message_broker.publish(MappingData(**MAPPING_DEFAULTS))
 
     def _on_combination_recorded(self, data: CombinationRecorded):
-        self.update_combination(data.combination)
+        combination = self._auto_use_as_analog(data.combination)
+        self.update_combination(combination)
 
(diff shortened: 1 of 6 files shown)


inputremapper/gui/controller.py
```diff
@@ -244,8 +244,36 @@ def copy_preset(self):
...
+                combination_list[i] = input_config.modify(analog_threshold=0)
+                new_combination = InputCombination(combination_list)
+                return new_combination
+
+        return combination
+
     def update_combination(self, combination: InputCombination):
         """Update the input_combination of the active mapping."""
+        combination = self._auto_use_as_analog(combination)
+
```
jonasBoss, line 276: This is kinda confusing. If I call update_combination with some new combination I would expect the combination to be the one I provided and not a modified version of it.

I think it is better if we call `_auto_use_as_analog` directly in the methods where we are sure we want to apply the logic. Currently this seems to be only in the `_on_combination_recorded` method.

jonasBoss (changes requested):

inputremapper/gui/controller.py
sezanzeb: Thanks

PR #628, opened 2023-02-21
Mapping filter control with dedicated Gtk.ListBox filter manager

ubunatic:
This is a followup on #621. It adds

* a generic list manager for Gtk.ListBoxes to show and hide rows
* a mapping filter control and the related UI elements

See the screens in #621

### DIFF
--- a/inputremapper/gui/components/gtkext/listbox_filter.py
+++ b/inputremapper/gui/components/gtkext/listbox_filter.py
@@ -33,6 +33,8 @@ class ListBoxFilter:
     >>> filter = ListBoxFilter(
     >>>     my_listbox,                  # Gtk.ListBox to be managed
     >>>     get_row_name=MyRow.get_name  # custom row name getter
+    >>>     filter_value="text"          # inital value
+    >>>     case_sensitive=True,         # override default:False
     >>> )
 
     To apply a filter use `set_filter` as follows.
@@ -53,8 +55,9 @@ class ListBoxFilter:
     ):
         self._controlled_listbox: Gtk.ListBox = listbox
         self._get_row_name = get_row_name or self.get_row_name
-        self._filter_value : str = filter_value
-        self._case_sensitive = case_sensitive
+        self._filter_value: str = ""
+        self._case_sensitive = False
+        self.set_filter(filter_value, case_sensitive=case_sensitive)
 
     @classmethod
     def get_row_name(T, row: Gtk.ListBoxRow) -> str:
--- a/inputremapper/gui/components/gtkext/listbox_filter.py
+++ b/inputremapper/gui/components/gtkext/listbox_filter.py
@@ -1,0 +1,130 @@
+# -*- coding: utf-8 -*-
+# input-remapper - GUI for device specific keyboard mappings
+# Copyright (C) 2023 sezanzeb <proxima@sezanzeb.de>
+#
+# This file is part of input-remapper.
+#
+# input-remapper is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# input-remapper is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with input-remapper.  If not, see <https://www.gnu.org/licenses/>.
+
+import gi
+
+from gi.repository import Gtk
+
+from typing import Iterator
+
+
+class ListBoxFilter:
+    """Implements UI-side filtering of list widgets.
+
+    The following example creates a new ``ListBoxFilter`` for a given ``Gtk.ListBox``.
+    It also sets all optional arguments to override some default behavior.
+
+    >>> filter = ListBoxFilter(
+    >>>     my_listbox,                  # Gtk.ListBox to be managed
+    >>>     get_row_name=MyRow.get_name  # custom row name getter
+    >>> )
+
+    To apply a filter use `set_filter` as follows.
+
+    >>> filter.set_filter("some text")
+    >>> filter.set_filter("More Text", case_sensitive=True)
+
+    """
+
+    MAX_WIDGET_TREE_TEXT_SEARCH_DEPTH = 10
+
+    def __init__(
+        self,
+        listbox: Gtk.ListBox,
+        get_row_name=None,
+        filter_value="",
+        case_sensitive=False,
+    ):
+        self._controlled_listbox: Gtk.ListBox = listbox
+        self._get_row_name = get_row_name or self.get_row_name
+        self._filter_value : str = filter_value
+        self._case_sensitive = case_sensitive
+
+    @classmethod
+    def get_row_name(T, row: Gtk.ListBoxRow) -> str:
+        """
+        Returns the visible text of a Gtk.ListBoxRow from both the row's `name`
+        attribute or the row's text in the UI.
+        """
+        text = getattr(row, "name", "")
+
+        # find and join all text in the ListBoxRow
+        text += " ".join(v for v in T.get_widget_tree_text(row) if v != "")
+
+        return text.strip()
+
+    @classmethod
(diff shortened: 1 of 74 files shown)


ubunatic:
Using `GtkSearchEntry` now.

P.S.: Ideally I would want to have sth. like this:
image

This is the VSCode search and I would love to have this on any list in any UI.

But one step at a time ;)

inputremapper/gui/components/gtkext/listbox_filter.py
```diff
@@ -0,0 +1,133 @@
...
+    @property
+    def filter_value(self):
+        return self._filter_value
+
+    @property
+    def case_sensitive(self):
+        return self._case_sensitive
+
+    # matches the current filter_value and filter_options with the given value
+    def match_filter(self, value: str):
```
sezanzeb, line 101: those should probably be docstrings instead

```suggestion
    def match_filter(self, value: str):
        """Match the current filter_value and filter_options with the given value."""
```

here and everywhere else

tests/integration/test_gui.py
```diff
@@ -304,6 +297,8 @@ def setUp(self):
         self.stop_injector_btn: Gtk.Button = get("stop_injection_preset_page")
         self.rename_btn: Gtk.Button = get("rename-button")
         self.rename_input: Gtk.Entry = get("preset_name_input")
+        self.mapping_filter_case_btn: Gtk.Button = get("mapping-filter-case-button")
```
sezanzeb, line 300:
```suggestion
        self.mapping_filter_case_btn: Gtk.ToggleButton = get("mapping-filter-case-button")
```

sezanzeb (commented):
Thanks a lot for your contributions! :)

Three points/thoughts from my side:

1. I would change the gui like this:

image

See https://github.com/sezanzeb/input-remapper/tree/628-gui-suggestion for the glade file that produces this

2. I would like the search to also search through the selected target, the programmed key/macro and searching for ABS_X should find analog output that is set to ABS_X. I would be totally fine with having this in a separate future PR.

3. Is the case sensitive search toggle really necessary? I'm not so sure that I would ever need a case-sensitive search here

ubunatic:
> 

1. ✅ merged your proposal, I was looking for this magic `linked` style but could find it. Thx! 
2. Separate PR is better. This one gets bigger and bigger
3. I would use it to search for generated text ("Alt L") vs. my own custom text ("ALT"). I have a lot of mappings now. And event the search is not enough actually. I would want a searchable tree or some other kind of grouping of things beyond being sorted A-Z. Maybe I would even group them manually. But for now the search is fine. Better than just scrolling. Let's start small.

sezanzeb:
allright

I found the "linked" class when I used the gtk inspector quod libet, it has a small dropdown button to the right of its search.

ubunatic:
I also added support for an explicit [GTK_PATH](https://docs.gtk.org/gtk3/running.html) in this PR now. This allows me to start the local development app with:
```
GTK_PATH=$PWD PYTHONPATH=$PWD bin/input-remapper-gtk -R
```
so that it will load the latest local files from `./data`.

I test my changes this way. First I start the app. Then I press ESC on the pkexec dialog. And then I try my changes in the GUI.

inputremapper/gui/components/gtkext/listbox_filter.py
```diff
@@ -0,0 +1,133 @@
...
+            return self._filter_value.lower() in value.lower()
+
+    # set and apply filter
+    def set_filter(self, filter_value: str, case_sensitive=False):
+        self._filter_value = str(filter_value)
+        self._case_sensitive = bool(case_sensitive)
+        self._gtk_apply_filter_to_listbox_children()
+
+    # apply filter to widget tree
+    def _gtk_apply_filter_to_listbox_children(self):
```
sezanzeb, line 120:
```suggestion
    def set_filter(self, filter_value: str, case_sensitive=False):
        """Set and apply filter."""
        self._filter_value = str(filter_value)
        self._case_sensitive = bool(case_sensitive)
        self._gtk_apply_filter_to_listbox_children()

    def _gtk_apply_filter_to_listbox_children(self):
        """Apply filter to widget tree."""
```

sezanzeb (approved):
Thanks. -R sounds handy for development.

Lets just move those two comments into docstrings.

Is this ready to merge then?

inputremapper/gui/reader_service.py
```diff
@@ -145,7 +145,11 @@ def pkexec_reader_service():
         exit_code = os.system(cmd)
 
         if exit_code != 0:
-            raise Exception(f"Failed to pkexec the reader-service, code {exit_code}")
+            ex = Exception(f"Failed to pkexec the reader-service, code {exit_code}")
+            if ingore_errors:
+                logger.warn(ex)
```
sezanzeb, line 150: Some people might have very permissive setups by being member of the "device" user group or something. In that case, it might make sense to at least try to run `input-remapper-control --command start-reader-service{debug}` without pkexec as a regular user.

sezanzeb (commented):
I want to make the 2.0 release soon.

Do you have any plans regarding the search looking through the outputs as well already? If not, I'd probably merge this after 2.0.

inputremapper/gui/reader_service.py
ubunatic: Actually, I would like to move better pkexec handling to another PR and actually avoid having a need for a `-R` flag (or similar).

[Here](inputremapper/gui/reader_service.py) is what I would be looking for in a later change. But this fails many tests, so I guess the story is bigger than I thought.

ubunatic:
> Do you have any plans regarding the search looking through the outputs as well already? If not, I'd probably merge this after 2.0.

Not sure this a good idea without more entry controls. The names in the list are defined by the input only; unless renamed manually. We would need more buttons, I guess:
image

But then the UI seems to get more complicated. The user would need to know that he is searching through text that is not visible in the list; only the first output will be shown on the right.   Let's have a discussion about better filters later.

If you want this in a 2.0, I would say we stay minimal with the change:
* remove the pkexec stuff (maybe replaced by development-only env var for now)
* keep the filter on visible input text only

I can do these changes tomorrow, after that I am off for the weekend and will also have less time next week.

inputremapper/gui/reader_service.py
sezanzeb: >  the "device" user group or something

I think it's called "input"

---

If the -R flag might be deprecated soon already, I wouldn't add it now

I think users being in the input group is the exception. I don't know right now how evdev behaves when an unprivileged user tries to use InputDevice objects and such. I guess you'd have to look out for any permission error and then exit. I would also use a special exit code for this >= 3.

ubunatic: OK, I will remove the `-R` now but will leave an undocumented [env var](https://github.com/ubunatic/input-remapper/pull/2/files#diff-8d46828515cda696c8e9c4ad14726b07a5a01f518519ab8def95e5b237175236R139) for development. As said, let's discuss better privilege handling separately.

jonasBoss:
> If you want this in a 2.0, I would say we stay minimal with the change:
> 
>   - remove the pkexec stuff (maybe replaced by development-only [env var](https://github.com/ubunatic/input-remapper/pull/2/files#diff-8d46828515cda696c8e9c4ad14726b07a5a01f518519ab8def95e5b237175236R139) for now)
>   - keep the filter on visible input text only

I think for now this makes the most sense. 

---
Searching through the whole mapping would require us to do the filtering in the data manager. I think that can be done by filtering inside the `publish_mapping()` method. The controller would set the filter in the data_manager, and then call `publish_mapping` whenever the filter changes.
I don't think we need more buttons for that we could just always search the whole mapping. I can not imagine that someone needs the extra granularity. If we want an advanced search, that should be done inside a dedicated menu. So we keep the main UI simple.

ubunatic:
> If we want an advanced search, that should be done inside a dedicated menu. So we keep the main UI simple.

agreed.

I think I am done with this PR. If there are smaller issues, I can address them next week. But I will not have time for major changes in the next days.

sezanzeb:
Finding something I'm looking for is easy when I can see it.

The output is not visible unless I click on a mapping.

Finding something that maps to a specific key is impossible unless I click on every mapping to look through every one of them.

Jetbrains IDEs look through everything when searching in the settings:

image

And I think this is the reason why the search works so good and is so helpful.

I think a search only makes sense when it also filters through the output. And I wouldn't make an extra window for it.

sezanzeb:
If you want to make an advanced search, then the advanced search would be tweakable to only look for specific things, while the regular search searches everything.

PR #632, opened 2023-02-23
Add service start and stop commands to setup script

ubunatic:
I think this is more convenient than browsing the shell history for `systemctl` commands, esp. for testing during development. Also, since the `start` command already loads the default bindings so that your keyboard is restored after testing something that required the system-level service to be stopped.

Usage:
```
scripts/setup.sh start  # start service and autoloads the default bindings 
scripts/setup.sh stop   # stops and disables service
```


sezanzeb:
Thanks

Can you please mention this and the shell script form the previous PR here: https://github.com/sezanzeb/input-remapper/blob/beta/readme/development.md ?

sezanzeb:
putting every paragraph on a single line isn't very ergonomic in IDEs by default:

image

ubunatic:
> putting every paragraph on a single line isn't very ergonomic in IDEs by default:
> 
> image
> 
> since the point of markdown is being readable without having to render it I'd rather have linebreaks in them.

Can we agree on a line length then?
The code look bad in my IDE.

sezanzeb:
> > putting every paragraph on a single line isn't very ergonomic in IDEs by default:
> > since the point of markdown is being readable without having to render it I'd rather have linebreaks in them.
> 
> Can we agree on a line length then? The code look bad in my IDE, when I had the code and preview in a split view.

Yeah, I'm a huge fan of the 88 characters of black.

sezanzeb:
Thanks

PR #635, opened 2023-02-25
Improved error formatting in status bar

### LINKED ISSUE #468: Unclear Error Messages (Beta)
"Remove the Analog Output Axis when specifying an macro or key output"

image

Is this error message related to this PR? I don't know what to do there

_Originally posted by @sezanzeb in https://github.com/sezanzeb/input-remapper/issues/442#issuecomment-1236126627_

sezanzeb:
closes #468

- Displays the error in the status bar directly, if there is only one
- the `__root__` substring is omitted from errors in the gui
- using gtks ellipsis functionality
- the error-icon was invisible when the application was loading
- the error lookup is not using redundant strings anymore
- checks for wrong targets in macro parameters
- Adds a helpful error message if the target is incorrect:

image

### DIFF
--- a/inputremapper/configs/mapping.py
+++ b/inputremapper/configs/mapping.py
@@ -118,8 +118,8 @@ def pydantify(error: type):
     """
     # See https://github.com/pydantic/pydantic/discussions/5112
     lower_classname = error.__name__.lower()
-    if lower_classname.endswith('error'):
-        return lower_classname[:-len('error')]
+    if lower_classname.endswith("error"):
+        return lower_classname[: -len("error")]
     return lower_classname
 
 
@@ -525,9 +525,7 @@ class Mapping(UIMapping):
         output_symbol = values.get("output_symbol")
 
         if not use_as_analog and not output_symbol and output_type != EV_KEY:
-            raise MissingMacroOrKeyError(
-                f'missing macro or key'
-            )
+            raise MissingMacroOrKeyError(f"missing macro or key")
 
         if (
             use_as_analog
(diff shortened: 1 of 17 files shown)


inputremapper/configs/mapping.py
```diff
@@ -337,28 +374,56 @@ def is_valid(self) -> bool:
...
+            return values
 
         if is_this_a_macro(symbol):
             try:
                 parse(symbol, verbose=False)  # raises MacroParsingError
-                return symbol
+                return values
             except MacroParsingError as exception:
                 # pydantic only catches ValueError, TypeError, and AssertionError
                 raise ValueError(exception) from exception
```
jonasBoss, line 401: Is this accounted for?

jonasBoss (approved):

inputremapper/configs/mapping.py
sezanzeb: Macro errors are displayed correctly and are readable to my satisfaction. Is that what you meant?

sezanzeb:
It will now also check for wrong targets in macro parameters

inputremapper/configs/mapping.py
jonasBoss: yeah, I see that you used custom errors everywhere else. So I was wondering why not here.

inputremapper/gui/controller.py
```diff
@@ -70,6 +78,19 @@
...
+    This does not include the base class name, which is transformed to snake case in
+    pydantic. Example pydantic error type: "value_error.foobar" for FooBarError.
+    """
+    # See https://github.com/pydantic/pydantic/discussions/5112
+    lower_classname = error.__name__.lower()
+    if lower_classname.endswith("error"):
+        return lower_classname[: -len("error")]
+    return lower_classname
+
+
```
jonasBoss, line 93: not sure if this belongs here, maybe move it to the validation_errors

jonasBoss (approved):

Issue #651, opened 2023-03-06
ImportError: cannot import name 'TypeAlias' from 'typing' (/usr/lib64/python3.9/typing.py)


theorin:
I use Almalinux 8.
After updating some packages input-remapper stopped working for me. 
I tried reinstalling the application several times, even rolling back to previous versions all the way to 1.4.1 but nothing worked.
I went back to the recent version (1.5.1), which gives me weird error when running ```input-remapper-gtk```:

```
[martin@hex ~]$ sudo pkill -f input-remapper && input-remapper-gtk -d
Traceback (most recent call last):
  File "/usr/bin/input-remapper-gtk", line 41, in <module>
    from inputremapper.configs.migrations import migrate
  File "/home/martin/.local/lib/python3.9/site-packages/inputremapper/configs/migrations.py", line 50, in <module>
    from inputremapper.configs.input_config import InputCombination, InputConfig
  File "/home/martin/.local/lib/python3.9/site-packages/inputremapper/configs/input_config.py", line 23, in <module>
    from typing import Tuple, Iterable, Union, List, Dict, Optional, Hashable, TypeAlias
ImportError: cannot import name 'TypeAlias' from 'typing' (/usr/lib64/python3.9/typing.py)
```

```input-remapper.service``` seems to be running
```
[martin@hex ~]$ systemctl status input-remapper.service
● input-remapper.service - Service to inject keycodes without the GUI application
   Loaded: loaded (/usr/lib/systemd/system/input-remapper.service; enabled; vendor preset: disabled)
   Active: active (running) since Mon 2023-03-06 16:00:28 CET; 9s ago
 Main PID: 7653 (input-remapper-)
    Tasks: 4 (limit: 203579)
   Memory: 24.4M
   CGroup: /system.slice/input-remapper.service
           ├─7653 /usr/bin/python3 /usr/bin/input-remapper-service
           └─7654 /usr/bin/python3 /usr/bin/input-remapper-service

Mar 06 16:00:28 hex systemd[1]: Starting Service to inject keycodes without the GUI application...
Mar 06 16:00:28 hex input-remapper-service[7653]: input-remapper-service 1.5.1 60541255c03bddfa68f275b97be57de641ceaeb5 https://github.c>
Mar 06 16:00:28 hex input-remapper-service[7653]: python-evdev 1.6.0
Mar 06 16:00:28 hex systemd[1]: Started Service to inject keycodes without the GUI application.
```

[...]

PR #652, opened 2023-03-06
Using NewType instead of TypeAlias

sezanzeb:
Closes https://github.com/sezanzeb/input-remapper/issues/651


theorin:
Thank you for a very quick commit. You are a lifesaver!
I have downloaded the latest version to try it out and everything works great.

Issue #656, opened 2023-03-13
Placeholder for the key-or-macro input


sezanzeb:
I think there is no placeholder text functionality for textviews (which is the kind of component that sourceview uses). So the editor component needs some extra logic to insert placeholder-text into itself, and to remove it when clicking into the textview. Furthermore, when the placeholder-text is written in it, it should just treat itself as empty and not emit any events.

I think this one is quite important.

https://github.com/sezanzeb/input-remapper/issues/649#issuecomment-1466482637

sezanzeb:
well, turns out if you set a placeholder text and then do set_text on the buffer, gtk will emit a "changed" event with an empty string before emitting another one with the target string for good measure. because why not.

sezanzeb:
For you to finish: https://github.com/sezanzeb/input-remapper/pull/658

PR #658, opened 2023-03-13
showing placeholder text

sezanzeb:
image

After clicking into it the text disappears.

This still needs tests, someone else do them please because https://github.com/sezanzeb/input-remapper/issues/656#issuecomment-1467029589 and I just can't deal with this any longer.

### DIFF
--- a/inputremapper/gui/components/editor.py
+++ b/inputremapper/gui/components/editor.py
@@ -402,6 +402,8 @@ class GdkEventRecorder:
 class CodeEditor:
     """The editor used to edit the output_symbol of the active_mapping."""
 
+    placeholder: str = _("Enter your output here")
+
     def __init__(
         self,
         message_broker: MessageBroker,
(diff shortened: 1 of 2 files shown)


sezanzeb:
Hint: if you return something else in `def code` like below:

```
    @property
    def code(self) -> str:
        """Get the user-defined macro code string."""
        if self._shows_placeholder():
            return "a"
```

the test_updates_mapping test suddenly works.

sezanzeb:
The test is possibly unrealistic, because I don't think there is a set_text call like that outside of tests, and the problem of having update_mapping called twice in the test might not actually happen during runtime.

sezanzeb:
Possible solution: add the listener after the focus-in event

sezanzeb:
I'll do the official 2.0.0 release after this

jonasBoss (approved):

PR #669, opened 2023-03-26
Fix `if_single` on Python 3.11

gebner:
With Python 3.11, I got the "Passing coroutines is forbidden, use tasks explicitly." error when using `if_single`.  This PR applies the same fix as in #555.

I've also added the missing `if`s when the branches are `None`.

### DIFF
--- a/inputremapper/injection/macros/macro.py
+++ b/inputremapper/injection/macros/macro.py
@@ -684,9 +684,11 @@ class Macro:
             self.context.listeners.remove(listener)
 
             if not listener_done.is_set() and self._trigger_release_event.is_set():
-                await then.run(handler)  # was trigger release
+                if then:
+                    await then.run(handler)  # was trigger release
             else:
-                await else_.run(handler)
+                if else_:
+                    await else_.run(handler)
 
         self.tasks.append(task)
 
--- a/inputremapper/injection/macros/macro.py
+++ b/inputremapper/injection/macros/macro.py
@@ -676,7 +676,7 @@ class Macro:
 
             resolved_timeout = _resolve(timeout, allowed_types=[int, float, None])
             await asyncio.wait(
-                [listener_done.wait(), self._trigger_release_event.wait()],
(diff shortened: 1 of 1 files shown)


inputremapper/injection/macros/macro.py
```diff
@@ -676,17 +676,19 @@ async def listener(event):
...
             self.context.listeners.remove(listener)
 
             if not listener_done.is_set() and self._trigger_release_event.is_set():
-                await then.run(handler)  # was trigger release
+                if then:
+                    await then.run(handler)  # was trigger release
             else:
-                await else_.run(handler)
+                if else_:
+                    await else_.run(handler)
```
sezanzeb, line 691:
```py
            if not listener_done.is_set() and self._trigger_release_event.is_set() and then:
                await then.run(handler)  # was trigger release
            elif else_:
                await else_.run(handler)
```

sezanzeb: however, this exceeds the maximum 88 chars of black, you'll have to extract `self._trigger_release_event.is_set()` and `not listener_done.is_set()` into variables with sensible names to make that line shorter.

sezanzeb:
@LunNova, if you have time, can you please look why pipelines are not working here? :)

jonasBoss:
I already approved the pipeline. But looking at github actions, they seem to be broken:
https://github.com/sezanzeb/input-remapper/actions/runs/4526050599
https://github.com/sezanzeb/input-remapper/actions/runs/4526050598
https://github.com/sezanzeb/input-remapper/actions/runs/4526050597
They all report "Queued" and I can't even cancel them or restart.

inputremapper/injection/macros/macro.py
sezanzeb: ah wait, this is probably wrong. It would always run else_ if then is not set.

sezanzeb:
@gebner, can you please repoen another pull request? Hopefully the pipeline will work then :)

jonasBoss:
I merged #683 since there was some wired bug with the workflows and black was complaining about the formatting.

PR #708, opened 2023-05-16
Add Brazilian Portuguese translation and update .po files

rffontenelle:
And before translating, updated pot file (using the documented commands) and other PO files (`msgmerge --update \<lang>.po input-remapper.pot`)


po/it_IT.po
```diff
@@ -16,188 +16,405 @@ msgstr ""
...
+msgstr "ferma l'iniezione"
+
+#: data/input-remapper.glade:201 data/input-remapper.glade:334
+msgid "Stop"
+msgstr "Ripristina impostazioni predefinite"
+
+#: inputremapper/gui/controller.py:710
+#, fuzzy
+msgid "Stopped the injection"
+msgstr "ferma l'iniezione"
```
sezanzeb, line 348: Thanks!

Do you know why `"Starting injection..."` and `"Stopped the injection"` have the same msgstr here?

rffontenelle: That's not harmful, so it is safe to leave as is.

When using `msgmerge` to update a PO file with an updated POT file, Gettext does a fuzzy match of the new strings against old strings in the PO file. When a fuzzy match is found, that new string receives the content of the previous plus a `#, fuzzy` tag.

That 'fuzzy' tag means the translation needs to be reviewed by a translator. Until then, that translation string will not show up in the translated UI.

For translators it is good because they might reuse it when updating the language translation. For the built software, no impact at all.

PR #711, opened 2023-05-19
Fix btn left confusion

jonasBoss:
#710 exposed an issue where an error would be logged, whenever a preset was applied due to the GUI check for `BTN_LEFT`. 
The check for `BTN_LEFT` was broken due to the introduction of `origin_hash`.

### DIFF
--- a/inputremapper/configs/preset.py
+++ b/inputremapper/configs/preset.py
@@ -232,15 +232,6 @@ class Preset(Generic[MappingModel]):
             existing = self._mappings.get(permutation)
             if existing is not None:
                 return existing
-
-        logger.error(
-            "Combination %s not found. Available: %s",
-            repr(combination),
-            list(
-                self._mappings.keys(),
-            ),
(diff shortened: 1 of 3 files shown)


inputremapper/configs/preset.py
```diff
@@ -232,21 +232,12 @@ def get_mapping(
...
-            ),
-        )
-
         return None
 
     def dangerously_mapped_btn_left(self) -> bool:
         """Return True if this mapping disables BTN_Left."""
-        if InputCombination([InputConfig.btn_left()]) not in [
-            m.input_combination for m in self
+        if InputConfig.btn_left().type_and_code not in [
```
sezanzeb, line 239: I still get the warning `WARNING input_config.py:268: No origin_hash set for {'type': 1, 'code': 272}` here

```suggestion
        if (ecodes.EV_KEY, ecodes.BTN_LEFT) not in [
```

Issue #716, opened 2023-05-26
Unable to apply or autoload any key mappings


alastair1616:
I was unable to apply or autoload on a fresh install of input-remapper on Arch Linux. Whenever I tried to apply a mapping I'd get the following error:
```
Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/inputremapper/gui/user_interface.py", line 256, in <lambda>
    "clicked", lambda *_: self.controller.start_injecting()
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/inputremapper/gui/controller.py", line 627, in start_injecting
    if self.data_manager.active_preset.dangerously_mapped_btn_left():
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/inputremapper/configs/preset.py", line 239, in dangerously_mapped_btn_left
    if (ecodes.EV_KEY, ecodes.BTN_LEFT) not in [
        ^^^^^^
NameError: name 'ecodes' is not defined
```
I was able to resolve the problem by manually adding:
`from evdev import ecodes`
to:
`/usr/lib/python3.11/site-packages/inputremapper/configs/preset.py`

arlarzelere3:
I had the same problem, and the same solution listed above fixed it.

Lcchy:
Same for me!

shawarden:
Same issue, same fix.

joaovitorpessoa:
Almost the same solution here...

I added `from evdev import ecodes` in `/usr/lib/python3/dist-packages/inputremapper/configs/preset.py`

PR #720, opened 2023-05-31
fix issue #716

Poldovico:
import a missing? dependency that made it impossible to apply presets in certain configurations


sezanzeb:
Thanks

Issue #723, opened 2023-06-08
ModuleNotFoundError: No module named 'inputremapper'; no devices detected


Shoxx98:
Traceback (most recent call last):
  File "/usr/bin/input-remapper-gtk", line 38, in <module>
    from inputremapper.gui.gettext import _, LOCALE_DIR
ModuleNotFoundError: No module named 'inputremapper'

Txordi00:
Same problem here. I think it started to fail after the update to gnome 44.1 (I am on Manjaro GNOME).

Shoxx98:
also noticed it after a big wave of updates.

sezanzeb:
> but no device is recognized so my remaps aren't working.

Is `input-remapper-control --list-devices` finding any?

sezanzeb:
> ModuleNotFoundError: No module named 'inputremapper'

Post the output of
- `find /usr/ -iname "*input*remapper*"`
- `python3 -c "import sys; print(sys.path)"`

Shoxx98:
input-remapper-control --list-devices                            ✔ 
Traceback (most recent call last):
  File "/usr/bin/input-remapper-control", line 30, in <module>
    from inputremapper.logger import logger, update_verbosity, log_info
ModuleNotFoundError: No module named 'inputremapper'


find /usr/ -iname "*input*remapper*"                           1 ✘ 
/usr/share/metainfo/io.github.sezanzeb.input_remapper.metainfo.xml
/usr/share/input-remapper
/usr/share/input-remapper/input-remapper.policy
/usr/share/input-remapper/input-remapper.svg
/usr/share/input-remapper/input-remapper-autoload.desktop
/usr/share/input-remapper/input-remapper.service
/usr/share/input-remapper/lang/sk/LC_MESSAGES/input-remapper.mo
/usr/share/input-remapper/lang/it_IT/LC_MESSAGES/input-remapper.mo
/usr/share/input-remapper/lang/fr_FR/LC_MESSAGES/input-remapper.mo
/usr/share/input-remapper/lang/it/LC_MESSAGES/input-remapper.mo
/usr/share/input-remapper/lang/uk_UA/LC_MESSAGES/input-remapper.mo
/usr/share/input-remapper/lang/fr/LC_MESSAGES/input-remapper.mo
/usr/share/input-remapper/lang/ru/LC_MESSAGES/input-remapper.mo
/usr/share/input-remapper/lang/zh/LC_MESSAGES/input-remapper.mo
[...]

PR #731, opened 2023-06-25
Using asyncio.run instead of loop.run_until_complete

sezanzeb:
Fixes https://github.com/sezanzeb/input-remapper/issues/723#issuecomment-1605930777


sezanzeb:
works with python 3.11 in manjaro and 3.10 in ubuntu

Issue #769, opened 2023-09-14
pydantic.errors.PydanticUserError


BillboardsAreAllLearing:
To help people understand your problems, run `sudo pkill -f input-remapper && input-remapper-gtk -d`, reproduce the problem and then share the output here:
```
Traceback (most recent call last):
  File "/usr/bin/input-remapper-control", line 31, in <module>
    from inputremapper.configs.migrations import migrate
  File "/usr/lib/python3.11/site-packages/inputremapper/configs/migrations.py", line 50, in <module>
    from inputremapper.configs.input_config import InputCombination, InputConfig
  File "/usr/lib/python3.11/site-packages/inputremapper/configs/input_config.py", line 51, in <module>
    class InputConfig(BaseModel):
  File "/usr/lib/python3.11/site-packages/inputremapper/configs/input_config.py", line 253, in InputConfig
    @root_validator
     ^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/pydantic/deprecated/class_validators.py", line 222, in root_validator
    return root_validator()(*__args)  # type: ignore
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/pydantic/deprecated/class_validators.py", line 228, in root_validator
    raise PydanticUserError(
pydantic.errors.PydanticUserError: If you use `@root_validator` with pre=False (the default) you MUST specify `skip_on_failure=True`. Note that `@root_validator` is deprecated and should be replaced with `@model_validator`.

For further information visit https://errors.pydantic.dev/2.3/u/root-validator-pre-skip
```

I obtain this error when starting input-remapper-control. I can't use as well ```input-remapper-control -h```. This is built from head on git via this AUR package on Arch Linux, [aur.archlinux.org/packages/input-remapper-git](https://aur.archlinux.org/packages/input-remapper-git).

PDXUser:
I have the same exact error:

```Traceback (most recent call last):
  File "/usr/bin/input-remapper-gtk", line 39, in <module>
    from inputremapper.gui.reader_service import ReaderService
  File "/usr/lib/python3.11/site-packages/inputremapper/gui/reader_service.py", line 55, in <module>
    from inputremapper.configs.input_config import InputCombination, InputConfig
  File "/usr/lib/python3.11/site-packages/inputremapper/configs/input_config.py", line 51, in <module>
    class InputConfig(BaseModel):
[...]

PR #773, opened 2023-09-18
Fix pydantic error (allow pydantic v1 and v2 at the same time)

stasadev:
## The Issue

* #769

## How This PR Solves The Issue

Input Remapper can continue to use pydantic v1 syntax because pydantic v2 allows it (see https://docs.pydantic.dev/2.3/migration/#continue-using-pydantic-v1-features)

I have added imports for pydantic v2 by default, and imports for pydantic v1 will be used as a fallback.

### DIFF
--- a/inputremapper/configs/input_config.py
+++ b/inputremapper/configs/input_config.py
@@ -25,7 +25,11 @@ from typing import Tuple, Iterable, Union, List, Dict, Optional, Hashable, NewTy
 from evdev import ecodes
 
 from inputremapper.input_event import InputEvent
-from pydantic import BaseModel, root_validator, validator
+
+try:
+    from pydantic.v1 import BaseModel, root_validator, validator
+except ImportError:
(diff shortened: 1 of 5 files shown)


sezanzeb:
Thanks. Tested it on ubuntu, still works. I'll just merge it, hopefully it works for everyone else as well.

Reviewdog isn't so fond of the imports, but I guess that's just what we are dealing with when writing python software that ends up in public repos.

sezanzeb:
Maybe a cleaner way might be to have our own module that does

```python
try:
    from pydantic.v1 import *
except ImportError:
    from pydantic import *
```

and then import our pydantic modules like

```python
from inputremapper.pydantic import BaseModel, root_validator, validator
```

but since this seems to be rather urgent, this is optional

bserem:
It also broke on debian testing, so a new release is probably a good idea.

The work done here fixes the problem. Thanks

PR #807, opened 2023-11-12
docs: remove redundant install step on Fedora

Merrit:
The `--now` flag to `systemctl enable` already starts the service immediately, no need to run a start command again.


sezanzeb:
Thank you

PR #841, opened 2024-01-26
Install icon to standard path, fix appstream-builder generation

joebonrichie:
appstream-builder was failing to generate the appstream metainfo from this package as it didn't provide an icon in a standard path.

appstream-builder states:
> Icons MUST be installed in /usr/share/pixmaps/*, /usr/share/icons/*,
> /usr/share/icons/hicolor/*/apps/*, or /usr/share/${app_name}/icons/*


sezanzeb:
Thank you!

PR #931, opened 2024-07-30
Add optional acceleration to the mouse macro.

jdoda:


inputremapper/injection/macros/macro.py
```diff
@@ -469,10 +469,11 @@ def add_event(self, type_: Union[str, int], code: Union[str, int], value: int):
         self.tasks.append(lambda handler: handler(type_, code, value))
         self.tasks.append(self._keycode_pause)
 
-    def add_mouse(self, direction: str, speed: int):
+    def add_mouse(self, direction: str, speed: int, accel: Optional[float] = None):
```
sezanzeb, line 472: I'd prefer `accel` to be written out as `acceleration`

inputremapper/injection/macros/macro.py
```diff
@@ -482,9 +483,19 @@ def add_mouse(self, direction: str, speed: int):
...
+                current_speed = 0.0
+            else:
+                current_speed = resolved_speed
+
             while self.is_holding():
-                handler(EV_REL, code, resolved_speed)
+                if resolved_accel and current_speed < resolved_speed:
+                    current_speed += resolved_accel
+
+                handler(EV_REL, code, value * int(current_speed))
```
sezanzeb, line 498:
```suggestion
                if resolved_accel and current_speed < resolved_speed:
                    current_speed += resolved_accel

                # To allow a macro of mouse(right, 1, 0.001) to move slowly, instead of
                # jumping to a speed of 1 eventually, progress the pointer by a pixel
                # by adding up the dither_remainder.
                dithered_speed = value * current_speed
                dither_remainder += dithered_speed - math.floor(dithered_speed)
                dithered_speed = math.floor(dithered_speed)

                if dither_remainder > 1:
                    dithered_speed += 1
                    dither_remainder -= 1

                handler(EV_REL, code, dithered_speed)
```

sezanzeb (commented):
Nicely done! Thanks

inputremapper/injection/macros/macro.py
sezanzeb: Oh, and initialize `dither_remainder = 0` above the while loop

sezanzeb: I'm not sure if dither is the correct term though. Dither seems to be specifically about avoiding banding using noise, but this isn't using noise.

sezanzeb:
Have you thought about the pros and cons of acceleration being a relative value, resulting in an acceleration curve instead of linear acceleration `(current_speed *= (1 + resolved_accel))`?

inputremapper/injection/macros/macro.py
jdoda: Done.

inputremapper/injection/macros/macro.py
jdoda: Yeah, that makes sense. I had been considering something like this, but wasn't sure how much difference it would make in practice. Since you've brought it up as well, I've added an implementation of this idea.

jdoda:
> Have you thought about the pros and cons of acceleration being a relative value, resulting in an acceleration curve instead of linear acceleration `current_speed *= 1 + resolved_accel`?

Yeah, I had considered whether a different acceleration curve would be better (and I've experimented with an exponential curve since you suggested it) and I like linear acceleration best. Most importantly, I think it just feels the best, but also it's easy to understand, easy to tune, and, even though this probably doesn't matter, it does actually match the usual physical definition of acceleration.

sezanzeb:
Once you pull the latest changes from the main branch, the pipeline will hopefully pass

inputremapper/injection/macros/macro.py
```diff
@@ -482,9 +483,25 @@ def add_mouse(self, direction: str, speed: int):
...
+                displacement = resolved_speed
+
             while self.is_holding():
-                handler(EV_REL, code, resolved_speed)
+                if resolved_accel and current_speed < resolved_speed:
+                    current_speed += resolved_accel
+                    current_speed = min(current_speed, resolved_speed)
+                    displacement_accumulator += current_speed
+                    displacement = int(displacement_accumulator)
+                    displacement_accumulator -= displacement
```
sezanzeb, line 502: Sorry, one more thing: I'd like to have an explanatory comment there why we are doing this, to avoid any future confusion.

```suggestion
                # Cursors can only move by integers. To get smooth acceleration for
                # small acceleration values, the cursor needs to move by a pixel every
                # few iterations. This can be achieved by remembering the decimal
                # places that were cast away, and using them for the next iteration.
                if resolved_accel and current_speed < resolved_speed:
                    current_speed += resolved_accel
                    current_speed = min(current_speed, resolved_speed)
                    displacement_accumulator += current_speed
                    displacement = int(displacement_accumulator)
                    displacement_accumulator -= displacement
```

inputremapper/injection/macros/macro.py
```diff
@@ -469,10 +469,11 @@ def add_event(self, type_: Union[str, int], code: Union[str, int], value: int):
         self.tasks.append(lambda handler: handler(type_, code, value))
         self.tasks.append(self._keycode_pause)
 
-    def add_mouse(self, direction: str, speed: int):
+    def add_mouse(self, direction: str, speed: int, acceleration: Optional[float] = None):
```
sezanzeb, line 472: To make `black` pass (1/2):

```suggestion
    def add_mouse(
        self,
        direction: str,
        speed: int,
        acceleration: Optional[float] = None,
    ):
```

To get `black` to put the parameter below each other, the trailing comma on the last parameter can be added.

tests/unit/test_macros.py
```diff
@@ -1012,6 +1012,17 @@ async def test_mouse(self):
...
+    async def test_mouse_accel(self):
+        macro_1 = parse("mouse(up, 10, 0.9)", self.context, DummyMapping)
+        macro_1.press_trigger()
+        asyncio.ensure_future(macro_1.run(self.handler))
+
+        sleep = 0.1
+        await asyncio.sleep(sleep)
+        self.assertTrue(macro_1.is_holding())
+        macro_1.release_trigger()
+        self.assertEqual([(2, 1, 0), (2, 1, -2), (2, 1, -3), (2, 1, -4), (2, 1, -4), (2, 1, -5)], self.result)
```
sezanzeb, line 1024: To make `black` pass (2/2):

```suggestion
        self.assertEqual(
            [(2, 1, 0), (2, 1, -2), (2, 1, -3), (2, 1, -4), (2, 1, -4), (2, 1, -5)],
            self.result,
        )
```

sezanzeb (commented):
To get the black pipeline to pass (https://github.com/sezanzeb/input-remapper/actions/runs/10295303648/job/28607445456?pr=931) some small changes need to be made. See below. Or run `black .` in the project directory to automatically format it for you.

PR #956, opened 2024-09-19
Do not exclude buttonless devices that have axes

StelardActek:
Changes to groups.py to enable devices such as racing pedals (which have multiple axes but no buttons) to be remapped in the GUI.

### DIFF
--- a/inputremapper/groups.py
+++ b/inputremapper/groups.py
@@ -389,11 +389,7 @@ class _FindGroups(threading.Thread):
             abs_capa = capabilities.get(EV_ABS)
             rel_capa = capabilities.get(EV_REL)
 
-            if (
-                key_capa is None
-                and abs_capa is None
-                and rel_capa is None
-            ):
+            if key_capa is None and abs_capa is None and rel_capa is None:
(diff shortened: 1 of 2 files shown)


StelardActek:
Hey @sezanzeb, I'll do my best but I'm not actually a python dev so I'm not really familiar with how these things are done.
A look at the output suggests one issue is just a formatting issue, easy enough to fix. Another seems to be failing because it is testing that ABS-only devices are excluded, which is what I'm intending to change here. If I can find that test I'll see about changing it. The last one... I'm not sure what that is about.

sezanzeb:
For the linter, you can get it to be formatted correctly quickly by doing

```
sudo apt install black
black input-remapper/inputremapper
```

(see https://github.com/psf/black)

There is luckily no need to do it by hand

jonasBoss:
From my side the change looks good!

sezanzeb:
I'll give it a try maybe tomorrow or something (just a quick test if gamepads still work and such, I don't have a foot pedal), and then merge it if there is no further objection.

sezanzeb:
Thank you!

PR #959, opened 2024-09-22
Escape portions of paths not intended to be globbed on

branan:
This fixes a crash with USB devices which contain glob-sensitive characters such as '*', '[]', or '{}' in their name.

### DIFF
--- a/inputremapper/gui/data_manager.py
+++ b/inputremapper/gui/data_manager.py
@@ -188,7 +188,7 @@ class DataManager:
         device_folder = get_preset_path(self.active_group.name)
         mkdir(device_folder)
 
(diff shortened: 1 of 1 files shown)


sezanzeb:
Thanks! Do you think you could provide a new test for this? Maybe it would be possible to copy a test in test_data_manager.py and insert some special characters in its name or something.

For the linting, you can automatically have your code be formatted to pass the pipeline. Just install black https://github.com/psf/black and run it on the code. For example

```
sudo apt install black
black input-remapper/inputremapper
```

sezanzeb:
merged in https://github.com/sezanzeb/input-remapper/pull/978

PR #960, opened 2024-09-28
Improved test setup

sezanzeb:
Applying a few things I learned in the past 2 years to input-remapper.

- Using patch.object instead of monkey-patches
- Moving stuff towards a dependency injection architecture, which allows better patches in tests. New classes, which might eventually be injected as dependencies:
  + PathUtils
  + UserUtils
  + Migrations
- Patching UInput class methods, instead of patching the UInput class as a whole
- Wrapped input-remapper-control in a class
- Moved static logging methods into logger class

With this, import order and patch order doesn't matter anymore

This allows:
- to make a decorator, that is applied to each test-class, to handily do the patches whenever we want them. To automatically and fully reset the patches after our tests, if we don't want them for whatever reason in other tests. And re-apply them arbitrarily later on. Using a decorator, because patches are commonly applied as a decorator already (`@patch.object(...)`) and it allows to do things without having to call super() in setUp and tearDown.
- to remove the old complicated test setup (which was, funnily, already a lot better than the older one)
- to more easily switch to pytest, if we want to.
- Tests didn't always work out-of-the-box in IDEs, now they do.
- using patch.object adds handy assertions that we can use on patched methods.


tests/integration/test_daemon.py
```diff
@@ -40,6 +39,8 @@ def gtk_iteration():
         Gtk.main_iteration()
 
 
+@setup_tests
+@setup_tests
```
sezanzeb, line 43: duplicate

tests/legacy_test_setup.py
```diff
@@ -24,9 +24,14 @@
 import os
```
sezanzeb, line 24: todo: remove file

tests/new_test.py
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/python3
```
sezanzeb, line 1: todo: rename to test_setup.py

tests/unit/test_control.py
```diff
@@ -99,12 +105,17 @@ def start_injecting(device: str, preset: str):
...
             daemon.injectors[device] = Injector()
 
-        daemon.start_injecting = start_injecting
+        patch.object(daemon, "start_injecting", start_injecting).start()
 
         global_config.set_autoload_preset(groups_[0].key, presets[0])
         global_config.set_autoload_preset(groups_[1].key, presets[1])
 
-        communicate(options("autoload", None, None, None, False, False, False), daemon)
+        self.input_remapper_control.communicate(
```
sezanzeb, line 113: todo: ditch the communicate method, call desired method directly (like .autoload, .start, etc.)

sezanzeb:
todo: add hint to developer guide that `import evdev; evdev.list_devices` should be used over `from evdev imoprt list_devices; list_devices`, so that patches in tests work. (except for the KEY_* stuff, no need to patch that stuff really)

sezanzeb:
todo:
- remove redundant quick_cleanup calls from test classes
- search for `original_*` variables and use pythons patch method instead

sezanzeb:
@jonasBoss this is quite a lot to review I guess, sorry. I'm not sure if there is anything worth looking at though, no features were added or bugs fixed, it's just refactoring and changes to tests. The `UserUtils` and `PathUtils` classes are a bit funky, but eventually I hope it might turn into a neat dependency-injected architecture, for which this is the first step.

sezanzeb:
A search for `^[a-z].+= ` in `inputremapper/` finds global things that should be injected dependencies instead.

sezanzeb:
image

sezanzeb:
@jonasBoss are you currently available?

jonasBoss:
yeah I currently have some free time

tests/integration/test_daemon.py
```diff
@@ -40,6 +40,8 @@ def gtk_iteration():
         Gtk.main_iteration()
 
 
+@test_setup
+@test_setup
```
jonasBoss, line 44: twice?

jonasBoss (approved):
So this is mostly about moving all the standalone functions into classes?

I don't see any obvious issues!

PR #964, opened 2024-10-02
Dependency injection

sezanzeb:
Blocked by https://github.com/sezanzeb/input-remapper/pull/960

- Further simplifies the test setup, because instead of having to clean up global singletons, they are created from scratch in each test to the testers hearts desire.
- Importing files doesn't have side-effects.
- Dependencies can be replaced by MagicMock objects, narrowing down the scope of unit-tests.

Not all classes are transitioned to this pattern. See comment below.


sezanzeb:
doing dependency injection for macros is rather difficult to do. In the end, the Mapping objects need the OutputParser (a class wrapping the `def parse` method and the others of that file) injected, which causes problems when serializing them. I haven't looked into it in more detail yet, as I didn't want to touch any actual logic right now to get it to work with serialization and deserialization.

But a little bit is done in this branch at least.

sezanzeb:
Thank you!

PR #965, opened 2024-10-04
Fixing release events when keys are released in order of pressing

sezanzeb:
### 1.

a + b -> c

1. press a
2. press b
3. release a
4. release b

the release event for a was never injected

---

### 2. 

"release_input" "on" released events, for which no key-down event was written before


inputremapper/injection/mapping_handlers/combination_handler.py
```diff
@@ -98,32 +104,35 @@ def notify(
...
 
         if is_activated:
             # send key up events to the forwarded uinput
             self.forward_release()
             event = event.modify(value=1)
-        else:
-            if self._output_state or self.mapping.is_axis_mapping():
-                # we ignore the suppress argument for release events
+
+        if not is_activated:
```
sezanzeb, line 133: This change may seem silly, but I honestly think the `if not` looks more readable than the `else` here. Same with is_pressed and is_released.

jonasBoss (changes requested):
I think there are still some bugs in the `notify()` method. There are 3 different states and 6 different state transitions:
 - `P` Passive
 - `A` Active
 - `S` Suppressed
 1. `P`->`P` with key down
 2. `P`->`P` with key up
 3. `P`->`A` key down
 4. `A`->`P` key up
 5. `P`->`S` key down
 6. `S`->`P` key up

I think the first 4 transitions are good, but not the transitions 5 and 6.

On transition 5 `self.forward_release()` will be executed (line 129). Basically releasing the input even though the mapping does not execute. A test for this would be two mappings:
`a+b+c->1` and `a+c->2` on mapping 1 set `release_combination_keys=False` and then input `a+b+c` the first mapping will be active, the second one suppressed but the key `a` is still released.

On transition 6 the line `return not self.should_release_event(event)` (line 125) is executed. Now there are two possibilities: the released key is the exact same key that triggered transition 5 -> Return False
the released key is a different one -> Return True

I think the second possibility can cause a stuck key like so (assuming transition 5 is fixed):
mappings: `a+b+d->1` and `c+d->2` 
press `a+b+c+d` -> mapping 1 is in sate `A` and 2 in state `S`
release `c` -> mapping 2 does transition 6 but the release of `c` is not forwarded because line 125 returns True (i think)

sezanzeb:
Thanks. I'm pretty impressed that you figured that out by looking at the code.

I reproduced transition 5 in `test_suppressed_doesnt_forward_releases` and fixed it.

I couldn't reproduce transition 6. Did I test that correctly in `test_no_stuck_key`?

jonasBoss:
I had to draw a state diagram in order to wrap my head around this and see all the possible state transitions.

Going to look at transition 6 again, as I could not check this without the fix for transition 5.

sezanzeb:
Sounds like something that I should do as well sometimes. I'm curious, could you please share the state diagram?

jonasBoss:
Bildschirmfoto_20241016_221743

jonasBoss:
I think I was wrong with the transition 6. `should_release_event` will always return True (assuming there are no duplicate key-up events). 

The only key that is not in `_requires_a_release` is the one that triggered transition 5, and that key must have triggered transition 3 in a mapping with higher priority. Therefore it did not get forwarded and cannot get stuck.

sezanzeb:
Ok, thank you. Shall we merge?

sezanzeb:
Should we keep `test_no_stuck_key`?

inputremapper/injection/mapping_handlers/combination_handler.py
```diff
@@ -133,18 +139,55 @@ def notify(
...
+
+            logger.debug("Sending %s to sub-handler", self.mapping.input_combination)
+            self._output_active = bool(event.value)
+            sub_handler_result = self._sub_handler.notify(event, source, suppress)
+
+            if is_pressed:
+                # Only if the sub-handler return False, we need a release-event later.
+                # If it handled the event, the user never sees this key-down event.
+                self.require_release_later(not sub_handler_result, event)
+                return sub_handler_result
```
jonasBoss, line 155: `is_activated` and `is_pressed` must be equal at this point.
`is_activated==False` and `is_pressd==True` can only happen on transition 1 which is already covered on line 116

inputremapper/injection/mapping_handlers/combination_handler.py
```diff
@@ -109,22 +111,26 @@ def notify(
...
         if is_activated:
-            # send key up events to the forwarded uinput
-            self.forward_release()
             event = event.modify(value=1)
         else:
-            if self._output_state or self.mapping.is_axis_mapping():
-                # we ignore the suppress argument for release events
+            # If not activated. All required keys are not yet pressed.
+            if self._output_active or self.mapping.is_axis_mapping():
+                # we ignore the `suppress` argument for release events
```
jonasBoss, line 133: this is always true because
`is_activated != self._output_active` and `is_activated == False`

jonasBoss (approved):
> Should we keep `test_no_stuck_key`?

I think it is fine to keep

> Ok, thank you. Shall we merge?

You can change the things I commented or not, I am not sure what is more readable

inputremapper/injection/mapping_handlers/combination_handler.py
jonasBoss: and `is_activated==True` and `is_pressed==False` is just impossible

sezanzeb:
I think we can remove the calls to `event.modify`

inputremapper/injection/mapping_handlers/combination_handler.py
sezanzeb: good catch

sezanzeb:
I split it into 4 separate methods
- _handle_freshly_activated
- _handle_freshly_deactivated
- _handle_still_activated
- _handle_still_not_activated

Is this more maintainable? I guess so

inputremapper/injection/mapping_handlers/combination_handler.py
```diff
@@ -109,74 +109,77 @@ def notify(
...
+        if changed:
+            if is_pressed:
+                return self._handle_freshly_activated(suppress, event, source)
+            else:
+                return self._handle_freshly_deactivated(event, source)
+        else:
+            if is_pressed:
+                return self._handle_no_change_press(event)
+            else:
+                return self._handle_no_change_release(event)
```
jonasBoss, line 123: I like the idea, but I think this is a bit misleading. If I read this correctly `_handle_no_change_release` will be called when the state changes from suppressed to passive. I guess a simple comment should be fine.

jonasBoss (approved):

inputremapper/injection/mapping_handlers/combination_handler.py
sezanzeb: I'm a bit lost, could you please propose a change with the comment?

sezanzeb:
Thank you so much @jonasBoss, after merging I realized I probably should have added you as a co-author of that commit. Great review.

sezanzeb:
I'll make a new 2.0.2 release in a few weeks

PR #966, opened 2024-10-04
Use packaging.version instead of pkg_resources

skitt:
pkg_resources is deprecated, this removes one set of uses.


sezanzeb:
Thank you so much!

Issue #972, opened 2024-10-12
How to write Macro in input


KAGEYAM4:
```
input-remapper 2.0.1 cf01699c5fba4d24bb81cce5c2f8bf024ffee265 https://github.com/sezanzeb/input-remapper
python-evdev 1.7.1
```

https://github.com/sezanzeb/input-remapper/blob/cf01699c5fba4d24bb81cce5c2f8bf024ffee265/readme/usage.md?plain=1#L18-L19
It states that we can use macro in input field. But it is not working. I am only able to use record button to fill the input field and then optionally rename the title of the input.

Also Macro in input, what would its behaviour would be like, do we macros exclusively for input ?

jonasBoss:
I guess the documentation is confusing. No there is no such thing as a macro as input. You can write macros into the **text-input-field** on the Output side.

KAGEYAM4:
okay. i will create a pr to fix this. Thanks

PR #973, opened 2024-10-12
Docs: clears misunderstanding on where Macros can be used

KAGEYAM4:
#972

### DIFF
--- a/readme/usage.md
+++ b/readme/usage.md
@@ -13,10 +13,10 @@ as shown in the screenshots below.
   <img src="usage_2.png"/>
 </p>
 
-In the text input field (right-side, output section), type the key to which you would like to map this input (left-side, input section).
+In the "Output" textbox on the right, type the key to which you would like to map this input.
 More information about the possible mappings can be found in 
 [examples.md](./examples.md) and [below](#key-names). You can also write your macro 
-into the text input field(right-side, output section). If you hit enter, it will switch to a multiline-editor with
+into the "Output" textbox. If you hit enter, it will switch to a multiline-editor with
 line-numbers.
 
 Changes are saved automatically. Press the "Apply" button to activate (inject) the 
--- a/readme/usage.md
+++ b/readme/usage.md
@@ -13,10 +13,10 @@ as shown in the screenshots below.
   <img src="usage_2.png"/>
 </p>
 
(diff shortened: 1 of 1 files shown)


readme/usage.md
```diff
@@ -13,10 +13,10 @@ as shown in the screenshots below.
   <img src="usage_2.png"/>
 </p>
 
-In the text input field, type the key to which you would like to map this input.
+In the text input field (right-side, output section), type the key to which you would like to map this input (left-side, input section).
 More information about the possible mappings can be found in 
 [examples.md](./examples.md) and [below](#key-names). You can also write your macro 
-into the text input field. If you hit enter, it will switch to a multiline-editor with
+into the text input field(right-side, output section). If you hit enter, it will switch to a multiline-editor with
```
sezanzeb, line 19:
```suggestion
into the "Output" textbox. If you hit enter, it will switch to a multiline-editor with
```

readme/usage.md
```diff
@@ -13,10 +13,10 @@ as shown in the screenshots below.
   <img src="usage_2.png"/>
 </p>
 
-In the text input field, type the key to which you would like to map this input.
+In the text input field (right-side, output section), type the key to which you would like to map this input (left-side, input section).
```
sezanzeb, line 16:
```suggestion
In the "Output" textbox on the right, type the key to which you would like to map this input.
```

sezanzeb (commented):
Good point. `into the text input field` was ambiguous.

What do you think about my suggested changes?

readme/usage.md
sezanzeb: (Please look at my other comment first, the order of my comments is mixed up)

readme/usage.md
KAGEYAM4: Sorry, it took some time. This is my first time going through this `Review-suggestion` workflow. So looked up some docs and videos.


I like your suggestion. How do i proceed, do i just use `Add suggestion to bach` and commit them? But this will create two commits on this pull request which is just a docs change. So i think one commit should be all we need. so do i pull, squash, force-push ?

sezanzeb: Don't worry, 2 days is certainly fast enough.

I'll squash the commits when I merge the PR, so there is nothing to do wrong

PR #982, opened 2024-10-22
Fix input-remapper-control for python < 3.12

DjUlt:
Had issue with input-remapper-control
Crashed with error ```unsupported operand type(s) for 'in': 'str' and 'EnumType'```
I added enum super class for Internals and Commands enum with 'in' implementation

### DIFF
--- a/bin/input-remapper-control
+++ b/bin/input-remapper-control
@@ -26,7 +26,7 @@ import logging
 import os
 import subprocess
 import sys
-from enum import Enum
+from enum import Enum, EnumMeta
 from typing import Union, Optional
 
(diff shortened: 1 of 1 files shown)


sezanzeb:
I see, testing `'autoload' in Commands` seems to be a feature of newer python versions. Thanks for pointing out.

I feel like a solution that requires less complicated engineering would be:

```py
from enum import Enum

class Commands(Enum):
    AUTOLOAD = "autoload"
    START = "start"
    STOP = "stop"
    STOP_ALL = "stop-all"
    HELLO = "hello"

print('autoload' in [command.value for command in Commands])  # True
print('asdf' in [command.value for command in Commands])  # False
```

PR #999, opened 2024-11-05
if_single code cleanup

sezanzeb:


sezanzeb:
included in https://github.com/sezanzeb/input-remapper/pull/1000

Issue #996, opened 2024-11-04
Home row mods issue when typing fast


Cash-coder:
### Expected Behavior:
- modify home row keys (H, J, K, L  &  G, F, D, S)  to add mods (Control, Super, Alt, Shift) when tapped or when hold
- If **tap** 'J' output = 'J'
- If **hold** 'J' output = something like `hold_keys(Control_L)`
  - this way **for example** the input: `hold key J + 1` = `Control + 1` 
  - input hold key K+U = Alt+U
  - if tap K = 'K'

### System notes
- Fedora release 40 (Forty)
- input remapper 2.0.1 + python-evdev 1.7.1
- Hyprland v0.44.1

### Attempted Solutions
- Solution 1: input `s` = `if_tap(key(s),  hold_keys(Control_L))`
  -**problem:** works well when holding, but not when tapping, probably due to a delay; for example typing 'sanctuary' outputs '**as**nctuary'
  - if I type **very slowly** then it works as expected, but I must type too slow
  - **subsolution:** changing the timeout variable: `	if_tap(key(s),  hold_keys(Control_L) **150**)` and many other values like 100, 200 ... 2000. All have the same behavior

- Solution 2: individual remaps, for example input `s+1` becomes `Control_L + 1`
  - like this `s+1 = key_down(Alt_L).key(1).key_up(Alt_L)`
  - **problem:** works very well, except it always inserts the key `s`. So something like `s+1` becomes `**s**+Control+1` instead of `Control+1`
  - **subsolution**: attempt to delete the inserted key with `KEY_BACKSPACE`. Like this: `key(KEY_BACKSPACE).key_down(Alt_L).key(1).key_up(Alt_L)` has the same behavior, the key `s` gets inserted anyway and it doesn't get deleted
  - I searched for an option that works similar to `prevent_default()` or `disable(s)`; without success

- other attempted solutions
  - `if_single(key(j), key_down(Super_L))` +  timeouts
  - weird chatGPT solution for the input `s+1` that has the same behavior: `modify("Alt_L", event("EV_KEY", "KEY_1", 1).event("EV_KEY", "KEY_1", 0))`

sezanzeb:
`if_tap(key(s),  hold_keys(Control_L))` will write the key once you release the key on your keyboard. Therefore typing will be weird.

`if_single(key(j), hold_keys(Super_L))` same with this.

This is the only way this can work, and it will always interfere with typing.

sezanzeb:
[...]

PR #1000, opened 2024-11-05
Add mod_tap macro

sezanzeb:
Adds a convenient macro for https://github.com/sezanzeb/input-remapper/issues/996

Name matches how other tools call it, like https://docs.qmk.fm/mod_tap

This differs from the second example of _Nested tap_ in https://github.com/qmk/qmk_firmware/blob/78a0adfbb4d2c4e12f93f2a62ded0020d406243e/docs/tap_hold.md#nested-tap-abba-nested-tap. There, it would write `ab` instead of `B`. Because I don't think there is any easy way to suppress an event that is not used as an input for something in input-remapper (b in this case). When it is seen by the mod_tap macro, the mod_tap macro has to decide what to do, right now, and it will have to inject its default key.

A `mode` parameter could be introduced later, which could accept something like `"QMK_DEFAULT"` or `"PERMISSIVE_HOLD"` and stuff.

Or, we could already maybe make `event` carry a `event.dont_forward` flag, that prevents it from being forwarded. In that case, the mod_tap macro has to inject all the events that arrived while the modifier was held down when it is released.

TODO:
- [ ] macros.md
- [ ] mention home-row-modifiers in examples.md
- [ ] tests based on the "default" column in https://github.com/qmk/qmk_firmware/blob/78a0adfbb4d2c4e12f93f2a62ded0020d406243e/docs/tap_hold.md#comparison-comparison
- [ ] https://zmk.dev/docs/keymaps/behaviors/mod-tap uses 200ms

### DIFF
--- a/inputremapper/injection/macros/macro.py
+++ b/inputremapper/injection/macros/macro.py
@@ -654,7 +654,7 @@ class Macro:
                     return
 
                 if event.value == 1:
-                    # Another key was pressed, trigger `else`
+                    # Another key was pressed
                     another_key_pressed_event.set()
                     return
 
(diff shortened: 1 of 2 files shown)


sezanzeb:
TODO: macros.md, examples.md, tests

sezanzeb:
I need someone who knows this stuff https://precondition.github.io/home-row-mods to look at this.

PR #1003, opened 2024-11-10
Better macro architecture

sezanzeb:
- The available arguments of macro functions (now called Tasks) are now very clearly defined, instead of using function-inspection and parameter-name conversions to do this stuff
- Each task has its own class. They can be extended in there without cluttering the Macro object anymore. There is space for more complex macros
- Task classes are very easy to understand now I think. New macros are easier to write.
- It uses the same class for both constant (KEY_A and stuff) and variable arguments (`$foo`), instead of wrapping the latter, while leaving the first one raw.

The new architecture is certainly bulkier than the old one with all the new classes, but I think it's worth it. New classes are: `Task`, `Argument`, `ArgumentConfig`, `RawValue`. `Variable` has been extended.

Renamed `MacroParsingError` to `MacroError` because this stuff is also flying around during runtime.


sezanzeb:
TODO test check_for_unknown_keyword_arguments

sezanzeb:
@jonasBoss no hard feelings if you are busy with other stuff. Maybe I could just merge this and the other PRs (https://github.com/sezanzeb/input-remapper/pull/1004 and https://github.com/sezanzeb/input-remapper/pull/1007) slowly one by one, and maybe in the future, if you feel like it, make a review and I'll fix any comments in a new PR.

Issue #1022, opened 2024-12-25
"The device was not grabbed" when remapping BTN_MIDDLE to BTN_LEFT


user529481:
Hello, I'm trying to remap the left button of my mouse (Evoluent VerticalMouse D) to the middle button for my arthritis. I'm getting "The device was not grabbed error" when I click "Apply". 

The program did allow me to register both the input and output correctly. I was even able to map the middle mouse button to a letter on the keypad successfully. Similarly I was able to map keyboard keys to other keys just fine. 

When I run `xinput list`
```
WARNING: running xinput against an Xwayland server. See the xinput man page for details.
⎡ Virtual core pointer                     id=2 [master pointer  (3)]
⎜   ↳ Virtual core XTEST pointer               id=4 [slave  pointer  (2)]
⎜   ↳ xwayland-pointer:15                     id=6 [slave  pointer  (2)]
⎜   ↳ xwayland-relative-pointer:15             id=7 [slave  pointer  (2)]
⎜   ↳ xwayland-pointer-gestures:15             id=8 [slave  pointer  (2)]
⎣ Virtual core keyboard                   id=3 [master keyboard (2)]
    ↳ Virtual core XTEST keyboard             id=5 [slave  keyboard (3)]
    ↳ xwayland-keyboard:15                     id=9 [slave  keyboard (3)]
```

sezanzeb:
Before you hit "Apply", is there some other error shown on the bottom of the window?

Zoldronesk:
Getting the exact same issue with a Logitech G502.

System:
  Kernel: 6.8.0-51-generic arch: x86_64 bits: 64 compiler: gcc v: 13.3.0 clocksource: tsc
  Desktop: Cinnamon v: 6.2.9 tk: GTK v: 3.24.41 wm: Muffin v: 6.2.0 vt: 7 dm: LightDM v: 1.30.0
    Distro: Linux Mint 22 Wilma base: Ubuntu 24.04 noble
Machine:
  Type: Desktop Mobo: INTEL model: B75 serial: <superuser required> uuid: <superuser required>
    UEFI: American Megatrends v: 4.6.5 date: 06/03/2021
CPU:
  Info: quad core model: Intel Core i5-3470 bits: 64 type: MCP smt: <unsupported> arch: Ivy Bridge
    rev: 9 cache: L1: 256 KiB L2: 1024 KiB L3: 6 MiB
[...]

PR #1024, opened 2024-12-27
Improved error status messages

sezanzeb:
https://github.com/sezanzeb/input-remapper/issues/1022 

- The error message when applying a preset failed was hiding the validation error previously.
- When the output type is incorrect for keys, it shows a better error message now.
- Switching to a different mapping within a preset shows the error of that mapping. Instead of always showing the error of the last mapping in the list.


sezanzeb:
TODO
- run local integration tests
- does this require updated translations?

Issue #941, opened 2024-08-17
mouse() does not chain with itself


umnikos:
While `mouse(left, 10)` and `mouse(up, 10)` both work, it's currently impossible to have a keyboard key that sends the mouse up and to the left. `mouse(left, 10).mouse(up, 10)` only moves the mouse left.

As there are only 4 such combined directions, an alternate way to have the same functionality would be to have mouse() accept 4 additional direction strings: `up_left`, `up_right`, `down_left`, `down_right`

sezanzeb:
I should have designed the `mouse` macro to allow something like `mouse(x=10, y=5)`.

You can, however, do something like

```
hold(
  event(EV_REL, REL_X, 1)
  .event(EV_REL, REL_Y, 2)
  .wait(10)
)
```

sezanzeb:
You could even make it accelerate like this:

```
set(speed, 0)
.hold(
  event(EV_REL, REL_X, $speed)
  .event(EV_REL, REL_Y, $speed)
  .add(speed, 1)
  .wait(20)
)
```

or

```
set(speed, 50)
.hold(
  event(EV_REL, REL_X, 1)
  .event(EV_REL, REL_Y, 2)
  .add(speed, -1)
  .wait($speed)
)
```

Though this isn't as smooth as the acceleration of the `mouse` macro.

sezanzeb:
You can now use `mouse_xy(10, 10)`

Or, you can do

```
parallel(
    mouse(up, 10),
    mouse(left, 10)
)
```

PR #1029, opened 2025-01-01
Add mouse_xy macro

sezanzeb:
https://github.com/sezanzeb/input-remapper/issues/941

- Allows moving into both x and y directions
- Turned acceleration into a fractional value, so that x and y accelerate proportionally. For consistency, this is changed in the original `mouse` macro as well. After the 2.1.0 release, this shouldn't change anymore.
- The injection of mouse events was 15 - 20% too slow


sezanzeb:
@jdoda what do you think?

jdoda:
> @jdoda what do you think?

Yeah, seems reasonable.

sezanzeb:
Thanks. I'll merge this in a week or something, after my other pull requests.

Issue #1046, opened 2025-01-29
Unable to bind BTN_SIDE + combos if BTN_SIDE + X/Y Axis is bound.


shawarden:
I have been trying to bind the back button `BTN_SIDE` + other mouse function such as Right button and X and Y movement.

If I have a combo of a given mouse button (`BTN_SIDE` for example) and both the `X` and `Y` axis bound to say a gamepad's analog axis I can't bind any additional that mouse button + another mouse button combos. If I have only 1 of the `X` or `Y` axis as a combo with that button set to an analog axis, I can bind any number of other button combos using the specific mouse button. When trying to add a button combo while having both X and Y axis bound, or trying to add the second axis while any other button combos are bound I get...

<details>
  <summary>input-remapper-gtk -d output</summary>
  
  ```python
14:35:13.955394 434041 service DEBUG combination_handler.py:62: {'input_combination': <InputCombination (<InputConfig (1, 276) BTN_EXTRA, None, c53313712b2a1f81f860e787177e9266, at 0x71e727b249b0>) at 0x71e7286e1720>, 'target_uinput': 'keyboard + mouse'}
14:35:13.955723 434041 service DEBUG combination_handler.py:62: {'input_combination': <InputCombination (<InputConfig (1, 275) BTN_SIDE, None, c53313712b2a1f81f860e787177e9266, at 0x71e727b77c00>, <InputConfig (2, 0) REL_X, None, c53313712b2a1f81f860e787177e9266, at 0x71e727b77250>) at 0x71e7286e15e0>, 'target_uinput': 'gamepad'}
14:35:13.955919 434041 service DEBUG combination_handler.py:62: {'input_combination': <InputCombination (<InputConfig (1, 275) BTN_SIDE, None, c53313712b2a1f81f860e787177e9266, at 0x71e727b77c00>, <InputConfig (2, 1) REL_Y, None, c53313712b2a1f81f860e787177e9266, at 0x71e727b77250>) at 0x71e7286e15e0>, 'target_uinput': 'gamepad'}
[...]

PR #1054, opened 2025-02-02
Don't suppress combinations that map to an analog axis

sezanzeb:
Fixes https://github.com/sezanzeb/input-remapper/issues/1046

1. BTN_EXTRA + X -> ABS_X
2. BTN_EXTRA + Y -> ABS_Y

`_create_hierarchy_handlers` has one key in `handlers`: BTN_EXTRA.

The analog input is omitted because of the `rank_by` method of the `CombinationHandler`.

This maps to a set containing the two combinations above. No HierarchyHandler is created, and both
combinations can be activated simultaneously.

3. BTN_EXTRA + BTN_LEFT -> a

As soon as this combination is added, the `_create_hierarchy_handlers` has two keys in handlers.

This causes `len(combinations_with_event)` to be > 1, and therefore creates a `HierarchyHandler`.

The `HierarchyHandler` suppresses combination 2.

I adjusted the `HierarchyHandler` to not suppress combinations 1 and 2.


shawarden:
After updating the issue no longer occurs. I can bind combos to 2 axis and additional buttons without errors.

PR #1064, opened 2025-02-15
Install dbus policy in /usr/share

skitt:
Since dbus 1.14.0, installing dbus policies in /etc is deprecated; see the deprecations section in the release announcement, https://lists.freedesktop.org/archives/dbus/2022-February/018131.html

dbus supports policies in /usr/share since 1.10, released in August 2015, so this should be safe.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -121,7 +121,7 @@ setup(
         ("/usr/share/icons/hicolor/scalable/apps/", ["data/input-remapper.svg"]),
         ("/usr/share/polkit-1/actions/", ["data/input-remapper.policy"]),
         ("/usr/lib/systemd/system", ["data/input-remapper.service"]),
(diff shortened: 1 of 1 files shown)


sezanzeb:
Thank you!

If I make a 2.1.1 release this weekend, is there a chance that it will be part of ubuntu 25.04?

skitt:
> If I make a 2.1.1 release this weekend, is there a chance that it will be part of ubuntu 25.04?

Yes — the Debian import freeze is on the 20th, so a release this weekend would have time to get in.

sezanzeb:
Thanks. Made the release

skitt:
And it’s on its way to Debian, in plenty of time for Ubuntu 25.04.

PR #1118, opened 2025-06-26
Update README.md

JosephScript:
Updated installation instructions with new AUR

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ sudo systemctl enable --now input-remapper
 ### Arch
 
 ```bash
-yay -S input-remapper-git
+yay -S input-remapper-bin
 sudo systemctl enable --now input-remapper
 ```
 


sezanzeb:
Hi. I haven't used anything arch based in quite some time. `input-remapper-git` is/was maintained by myself. I have a few questions:
- Is `input-remapper-git` broken? There was a comment in may that mentions import problems
- The version of `input-remapper-git` seems to be old, but since it builds from git it should be fine. I guess I should update the version number in the pkgbuild file. Is the version number the reason for this PR?
- Should I be concerned about mentioning an AUR package that I don't maintain myself in the readme?

sezanzeb:
I just updated the AUR package with a new version string :)

> If it's possible to update the package I'm happy to close the issue.

closing it then

PR #1150, opened 2025-08-23
Added automatic stopping and re-starting of injection when recording …

sabudum:
…input.

When hitting the record button to grab inputs for remapping. If the preset is currently running, the program will stop it temporarily until recording finishes, then re-start the injection afterwards.

### DIFF
--- a/inputremapper/gui/controller.py
+++ b/inputremapper/gui/controller.py
@@ -615,12 +615,12 @@ class Controller:
 
         Updates the active_mapping.input_combination with the recorded events.
         """
+        previous_preset = ""
         state = self.data_manager.get_state()
+
         if state == InjectorState.RUNNING or state == InjectorState.STARTING:
-            self.data_manager.stop_combination_recording()
-            self.message_broker.signal(MessageType.recording_finished)
-            self.show_status(CTX_ERROR, _('Use "Stop" to stop before editing'))
-            return
+            previous_preset = self.data_manager.active_preset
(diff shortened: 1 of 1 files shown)


sabudum:
> Thanks, seems to work well!
> 
> After recording something, the "Record" button greys out:
> 
> image
> It shouldn't do that anymore then, right?

Tried to solve that, but couldn't find the cause.
You can select a different mapping and the button un-greys.

sezanzeb:
Probaby some usage of `set_sensitive` controls the recording button. https://lazka.github.io/pgi-docs/#Gtk-4.0/classes/Widget.html#Gtk.Widget.set_sensitive

sabudum:
> Probaby some usage of `set_sensitive` controls the recording button. https://lazka.github.io/pgi-docs/#Gtk-4.0/classes/Widget.html#Gtk.Widget.set_sensitive

Figured it would be, but could not find the code where it is being set.

sabudum:
> I this case you could add a print statement above all the calls to `set_sensitive` and then run the application. The information printed to the console could help you to figure out which call it is.
> 
> Once you figured that out, you can use `traceback.print_stack()` to understand how it is called (Or follow the usage of methods and classes through the code).
> 
> Or use a debugger and breakpoints.

I've traced all calls to set_sensitive()
There's only one call that disables the Record button.
And it is called when no active mapping is selected. My guess is that when the injection starts it clears the selected mapping, but also couldn't find how to solve that.

PR #1183, opened 2025-10-18
add toggle macros

Renari:
Adds the ability to play a macro indefinitely until the activation button is pressed again.

### DIFF
--- a/tests/unit/test_macros/test_toggle.py
+++ b/tests/unit/test_macros/test_toggle.py
@@ -36,25 +36,34 @@ class TestToggle(MacroTestBase):
     async def test_toggle(self):
         # repeats key(a) as long as macro is toggled
         macro = Parser.parse("toggle(key(a))", self.context, DummyMapping)
-
-        macro.press_trigger()
-        macro.release_trigger()
-        asyncio.ensure_future(macro.run(self.handler))
-        await asyncio.sleep(0.2)
-        self.assertGreater(len(self.result), 2)
-
         code_a = keyboard_layout.get("a")
-        self.assertGreater(self.result.count((EV_KEY, code_a, 1)), 2)
 
         self.assertEqual(self.count_child_macros(macro), 1)
         self.assertEqual(self.count_tasks(macro), 2)
 
+        # Start
         macro.press_trigger()
         macro.release_trigger()
-        result_count = self.result.count((EV_KEY, code_a, 1))
-        await asyncio.sleep(0.2)
+        asyncio.ensure_future(macro.run(self.handler))
+
+        await asyncio.sleep(0.1)
+        count_1 = self.result.count((EV_KEY, code_a, 1))
+        self.assertGreater(count_1, 2)
+
+        # Keeps writing more
+        await asyncio.sleep(0.1)
+        count_2 = self.result.count((EV_KEY, code_a, 1))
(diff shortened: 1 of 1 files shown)


Renari:
I'm not sure why but even adding a 10ms delay this immediately starts destroying any program that obtains focus, including my desktop forcing me to reboot.  This behaves very similarly to `repeat` which works perfectly fine.

Renari:
This should be working now, using `release_trigger` to track the 2nd key press is functioning.

inputremapper/injection/macros/tasks/toggle.py
```diff
@@ -0,0 +1,59 @@
...
+        self.release_count += 1
+
+    async def run(self, callback) -> None:
+        macro = self.get_argument("macro").get_value()
+        macro_has_wait = any(isinstance(task, WaitTask) for task in macro.tasks)
+
+        while self.release_count <= 1:
+            await macro.run(callback)
+            if not macro_has_wait:
+                await asyncio.sleep(0.1)
```
sezanzeb: `hold` and `repeat` aren't doing this. And what if someone needs this to be as fast as possible.

However, `hold` does not freeze. `hold` always adds 1ms of sleep, which is already enough to keep things running smoothly.

I'd suggest we keep the behavior as consistent as possible across all repeating macros.

sezanzeb:
(Tests on github should run now without me having to approve them each time)

tests/unit/test_macros/test_toggle.py
```diff
@@ -0,0 +1,61 @@
...
+
+        self.assertEqual(self.count_child_macros(macro), 1)
+        self.assertEqual(self.count_tasks(macro), 2)
+
+        macro.press_trigger()
+        macro.release_trigger()
+        result_count = self.result.count((EV_KEY, code_a, 1))
+        await asyncio.sleep(0.2)
+        # ensure that the macro has stopped
+        self.assertEqual(result_count, self.result.count((EV_KEY, code_a, 1)))
```
sezanzeb: Thank you. How about

```suggestion
@test_setup
class TestToggle(MacroTestBase):
    async def test_toggle(self):
        # repeats key(a) as long as macro is toggled
        macro = Parser.parse("toggle(key(a))", self.context, DummyMapping)
        code_a = keyboard_layout.get("a")

        self.assertEqual(self.count_child_macros(macro), 1)
        self.assertEqual(self.count_tasks(macro), 2)

        # Start
        macro.press_trigger()
        macro.release_trigger()
        asyncio.ensure_future(macro.run(self.handler))

        await asyncio.sleep(0.1)
        count_1 = self.result.count((EV_KEY, code_a, 1))
        self.assertGreater(count_1, 2)

        # Keeps writing more
        await asyncio.sleep(0.1)
        count_2 = self.result.count((EV_KEY, code_a, 1))
        self.assertGreater(count_2, count_1)

        # Stop
        macro.press_trigger()
        macro.release_trigger()
        result_count = self.result.count((EV_KEY, code_a, 1))
        await asyncio.sleep(0.2)
        # ensure that the macro has stopped
        self.assertEqual(result_count, self.result.count((EV_KEY, code_a, 1)))
```

This tests if it keeps writing more events as time passes

Renari:
> Is this ready to be merged now?

yep

PR #1197, opened 2025-11-01
Pydantic v2

sezanzeb:
v1 is not completely compatible with 3.14, and might break soon.

It seems for now this is not critical for input-remapper. Fedora users have reported that it works with that other recent fix.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,6 +21,7 @@ jobs:
         # cache-dependency-path: setup.py
     - name: Install dependencies
       run: |
+        lsb_release -a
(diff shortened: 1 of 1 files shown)


sezanzeb:
python3-pydantic-core is not available in ubuntu 24.04 LTS.

24.04 still has pydantic v1.

I don't think I want to stop supporting 24.04 yet, and therefore want to delay the switch to pydantic v2.

Python 3.15 will be released on 2026-10-01 or something (https://peps.python.org/pep-0790/), so in about a year. It remains to be seen if pydantic v1 breaks with that release for input-remapper.

PR #1211, opened 2025-11-15
Fix XDG_CONFIG_HOME being ignored

### LINKED ISSUE #448: Custom config directory

Jorenar:
Fixes #448
Closes #595

The location `~/.config` is a fallback to `$XDG_CONFIG_HOME` if the environment variable is not set,
as per [XDG Base Directory specification](https://specifications.freedesktop.org/basedir/latest/).

### DIFF
--- a/inputremapper/configs/paths.py
+++ b/inputremapper/configs/paths.py
@@ -32,14 +32,15 @@ from inputremapper.user import UserUtils
 
 # TODO maybe this could become, idk, ConfigService and PresetService
 class PathUtils:
-    rel_path = ".config/input-remapper-2"
-
     @staticmethod
     def config_path() -> str:
-        # TODO when proper DI is being done, construct PathUtils and configure it in
-        #  the constructor. Then there is no need to recompute the config_path
+        # TODO when proper DI is being done, construct PathUtils and configure it
+        #  in the constructor. Then there is no need to recompute the config_path
         #  each time. Tests might have overwritten UserUtils.home.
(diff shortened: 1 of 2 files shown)


sezanzeb:
Thanks, looks good! In my case $XDG_CONFIG_HOME is not set.

test_migration probably fails because `patch_paths` in `patches.py` doesn't work as expected anymore. Could you please try to fix it?

sezanzeb:
In any case, overwriting the XDG_CONFIG_HOME variable for tests would allow more test coverage by requiring less patches

Jorenar:
Okay, unexpectedly, looks like the `$XDG_CONFIG_HOME` variable is actually set in the runner!

sezanzeb:
Thank you! I actually didn't know about patch.dict, looks very handy

PR #1216, opened 2025-11-22
Use raw string literals for regexes

skitt:
This avoids confusion between Python escapes and regex escapes.

Reported in https://bugs.debian.org/1066817


sezanzeb:
Thanks!

Issue #1170, opened 2025-09-27
setup.py and especially data_files are deprecated


SammysHP:
```
        ********************************************************************************
        Please avoid running ``setup.py`` directly.
        Instead, use pypa/build, pypa/installer or other
        standards-based tools.

        By 2025-Oct-31, you need to update your project and remove deprecated calls
        or your builds will no longer be supported.

        See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html for details.
        ********************************************************************************
```

There is a workaround using pypa-build and pypa-installer, but `data_files` isn't supported anymore, so everything will be installed in `/usr/lib/python/site-packages`.

Please update to a non-deprecated build system or package maintainers will have a hard time packaging input-remapper.

SammysHP:
I've stopped giving myself headaches with Python setup tools a long time ago.

These might be helpful resources:

https://discuss.python.org/t/should-there-be-a-new-standard-for-installing-arbitrary-data-files/7853  
https://github.com/pypa/setuptools/discussions/2648

In theory it would make sense to copy these files manually to the target locations, but right now, python-installer installs them to `site-packages` as well, which clutters that directory. Additionally, with so many files in various system directories, a configurable Makefile like generated by CMake might be helpful. I guess there's something similar for Python.

Btw here are the changes for the Arch Linux PKGBUILD if going the path through pypa-build → wheel → pypa-installer:

```sh
…

makedepends=('git' 'python-build' 'python-installer' 'python-wheel' 'python-setuptools')

…

build() {
	cd "$srcdir/$_pkgname"
	python -m build --wheel --no-isolation
}

package() {
	cd "$srcdir/$_pkgname"
	python -m installer --destdir="$pkgdir" dist/*.whl
}
```

PR #1221, opened 2025-11-29
pyproject.toml, new build tool, improve github workflows

sezanzeb:
closes https://github.com/sezanzeb/input-remapper/issues/1170

pip, in many cases, fails to install data files, which need to go into system paths, and instead puts them (despite them being absolute paths, and common workarounds/fixes) into /usr/lib/python3/inputremapper/usr/share/...

python3 setup.py install is deprecated. So I had to remove the existing installation shell script.

meson fails to install the module into a path that can actually be imported, and its python features require one to specify each individual file of the module.

So instead input-remapper will use a custom python solution. Hopefulls this works well enough to prevent all ModuleNotFoundErrors in the future.

Surprisingly, this is not much more code that would otherwise be required. We already had the data_files list and would need it with meson as well with more boilerplate. The python code to generate translations was already there. We (even with meson) need to figure out the correct python-package target path (see code comments). And on the upside, we have a lot of flexibility and control without having to dig into configuration, unexpected behavior and deal with any further limitations.

This PR also changes github workflows
- runs tests not as root anymore, which previously probably made them run with ubuntu-lts python3.12 every time, instead of the min and max versions
- adds ruff
- tests if the .deb build and the install module install correctly and result in a working system

# Important changes for packaging:

- setup.py is replaced with pyproject.toml now. So `setup.py install` cannot be used anymore
- building can be done via `python3 -m install --root ./foo`, which puts all files in the expected target directory structure in root. In my case, on ubuntu 25.04, the module ends up in `foo/usr/lib/python3/dist-packages`. It picks the best option from `python3 -c "import sys; print(sys.path)"` for this. If this is missing a feature to make it helpful for packaging, please let me know
- `commit_hash.py` is not built anymore, instead, values are inserted into `installation_info.py`

### DIFF
--- a/.github/workflows/installation.yml
+++ b/.github/workflows/installation.yml
@@ -57,7 +57,7 @@ jobs:
       run: |
         cd $GITHUB_WORKSPACE
 
-        yes | sudo python3 -m install.uninstall
+        printf y | sudo python3 -m install.uninstall
 
         if pip show input-remapper; then
             echo "Expected input-remapper package to be uninstalled"


sezanzeb:
TODO:
- [x] Fix ci
- [x] backup my computer
- [x] try `python3 -m install --root /` locally (works fine in the workflow vm already)
- [x] wait until next weekend in case something comes to mind
- [x] Tell package maintainers about the change
- [x] shell.nix uses setup.py
- [ ] Merge

SammysHP:
I don't have much experience with Python packaging, so can't say if it's good or not.

sezanzeb:
Updated arch package (will need to use correct pkgver once this change is merged into main):

- python3-pip is required in makedepends
- removed build step because the install module installs directly
- removed setuptools

```bash
# Maintainer: sezanzeb b8x45ygc9@mozmail.com
# Contributor: Sven Karsten Greiner <sven@sammyshp.de>

pkgname=input-remapper-git
_pkgname=${pkgname%-git}
pkgver=2.2.0.r105.g374dcccf
pkgrel=1
pkgdesc="A tool to change and program the mapping of your input device buttons."
arch=('any')
url="https://github.com/sezanzeb/input-remapper/"
license=('GPL-3.0-or-later')
depends=('gtk3' 'gtksourceview4' 'python-cairo' 'python-evdev' 'python-gobject' 'python-packaging' 'python-psutil' 'python-pydantic' 'python-dasbus')
optdepends=('xorg-xmodmap: symbol names matching the keyboard layout')
makedepends=('git' 'python-pip')
provides=("$_pkgname")
conflicts=("$_pkgname" 'key-mapper')
source=("$_pkgname::git+https://github.com/sezanzeb/input-remapper.git")
sha256sums=('SKIP')

pkgver() {
	cd "$srcdir/$_pkgname"
	git describe --long --tags | sed 's/\([^-]*-g\)/r\1/;s/-/./g'
}

package() {
	cd "$srcdir/$_pkgname"
	python3 -m install --root $pkgdir
}
```

skitt:
> I haven’t checked this PR specifically, but I maintain other Python projects that have switched to `pyproject.toml` without `setup.py` without difficulty, so this should be fine.

It turns out that the installation module can’t be used, at least for Debian and derivatives, since it relies on `pip` to actually install the module. For the time being I’ve [replicate the work the module does](https://salsa.debian.org/python-team/packages/input-remapper/-/commit/471c1f439c961d54c857624cec1148beb972a4b6), which actually means the package needs fewer patches! But it might make things simpler for packagers if the installation module had a mode where it does everything *apart* from installing the module itself — at least in Debian, the packaging tools know how to install Python modules in the right place, it’s everything else that needs to be done manually (install the binaries, language files etc.).

PR #1233, opened 2026-01-07
Install only some components

sezanzeb:
Addresses https://github.com/sezanzeb/input-remapper/pull/1221#issuecomment-3716535241

### DIFF
--- a/.github/workflows/installation.yml
+++ b/.github/workflows/installation.yml
@@ -78,3 +78,42 @@ jobs:
             echo "Expected input-remapper service to be missing"
             exit 1
         fi
+    - name: Partial generic installation
+      run: |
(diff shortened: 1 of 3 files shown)


sezanzeb:
@skitt With this PR you could do `python3 -m install --root foo --components data_files` to only install data files. Is this an adequate solution?

If yes, I'll merge it right away

sezanzeb:
data_files also builds the language. I was thinking about separating them, or maybe even having language, udev, gui, systemd, etc. components to choose from, but that seemed overkill

skitt:
This is great, thanks! Let’s keep it simple :wink: — I don’t think splitting it further would be all that useful.

Issue #893, opened 2024-05-18
Mapping a combo involving the dpad disables the dpad for the original controller


zany130:
### Discussed in https://github.com/sezanzeb/input-remapper/discussions/890

<div type='discussions-op-text'>

<sup>Originally posted by **zany130** May 12, 2024</sup>
I am trying to map  hotkeys on the input -remapper controller on top of the original controller for example: `start + dpad up` to be `vol up` `start + dpad down` be ` vol down` etc

but each of the buttons when pressed by themselves should come from the original controller

this, for some reason, disables the dpad in the original controller. Is this intentional?

no other button exhibits this behavior. For example, I have `start + y` set to `alt+f4`, and both start and y work normally when pressed by themselves. meaning they come from the original non emulated controller

I can create a dpad mapping in the emulated controller and that will work but I want all normal gamepad inputs to come from the original controller

bug? @sezanzeb  </div>

the creating dpad mapping in the emulated controller no longer works. If I make any mapping involving the dpad. the dpad no longer works at all while that profile is in effect

https://gist.github.com/zany130/301c548addc078bc7d32111babcb1d6d

amigthea:
I can confirm the behaviour

alxns7:
I have something similar, where mapped dpad buttons stop working on the original controller when the injection is applied.

For example, I have `START + DPad-X Left` mapped to `Super_L + Shift_L + Left` and same for `Right` (to move windows around displays)

I have Steam big picture running, the dpad works normally. When I apply the injection, `dpad right/left` stop working, but `dpad up/down` still work, while I can move my windows left/right.

As soon as I stop the injection, `dpad left/right` work again. It's really odd.

sezanzeb:
Please install from the "fix_hat_mapping" branch (https://github.com/sezanzeb/input-remapper/pull/1274) and give feedback if your problems are fixed

sezanzeb:
Merged fixes to main

PR #1274, opened 2026-03-22
Gamepad fixes

### LINKED ISSUE #1152: Stuff broken when mapping gamepads
Don't remember the details though. Need to play around a bit. I think it was something obvious. And the UI sometimes defaults to "Analog Axis", I think something confused me about that.

There was something about a combination being detected as already-existing, even though it doesn't. Maybe involving an analog axis.

This issue is sort of a reminder for later this year

branch is "gamepad_fixes"

sezanzeb:
closes #1152 and https://github.com/sezanzeb/input-remapper/issues/893

Fixes:
- Mapping a combination onto dpad-right keeps an uncombined dpad-left working now
- Default deadzone for mapping joysticks to keys (analog_threshold) is 30 now, instead of 1. This makes recording joystick movement a lot easier, because it won't record both ABS_X and ABS_Y all the time for tiny movements
- If only one direction of a joystick or the dpad was mapped, the other direction was not forwarded
- When mapping triggers to keys, they are now correctly counted as released when not pressed, instead of a negative joystick direction

Quality:
- Added mechanism that logs EV_ABS and EV_REL events without flooding the terminal
- Added test for gamepads with a joystick abs range of 0 to +256 (instead of -32768 to +32768)
- Handlers are not overwriting event values anymore. They set an attribute instead for on/off ("pressed")
- Split test_event_pipeline file to multiple with one class per file
- Comments and cleanup
- Added more typehints, fixed a few mypy errors
- Stricter typing for the methods and functions that print the structure of the MappingHandlers

TODO:
- [x] check if using EV_REL as input still works

Automated tests for the:
- [x] unmapped direction of a joystick fix
- [x] dbad combination fix
- [x] trigger fix


inputremapper/injection/mapping_handlers/abs_to_btn_handler.py
```diff
@@ -94,29 +97,38 @@ def notify(
...
+        is_positive = value > mid_point
+        is_negative = value < mid_point
+        is_release = value == 0
+        positive_mapped = threshold > mid_point
+        negative_mapped = threshold < mid_point
+
+        if positive_mapped and is_positive:
+            pressed = positive_mapped and (value >= threshold)
+        elif negative_mapped and is_negative:
+            pressed = negative_mapped and (value <= threshold)
```
sezanzeb:
```suggestion
        if positive_mapped and is_positive:
            pressed = value >= threshold
        elif negative_mapped and is_negative:
            pressed = value <= threshold
```

inputremapper/injection/mapping_handlers/axis_switch_handler.py
```diff
@@ -39,6 +39,9 @@
 class AxisSwitchHandler(MappingHandler):
     """Enables or disables an axis.
 
+    This is used when a combination involving an analog input (rel or abs) is mapped
+    to another analog input (rel or abs). I think.
```
sezanzeb:
```suggestion
    to another analog output (rel or abs). I think.
```

PR #1276, opened 2026-03-26
Fix memory leak in event_reader; autoload presets for all device instances

nevion:
## Summary

- **Fix memory leak in EventReader**: `read_loop` created a new `asyncio.Task` on every input event to race `events_ready.wait()` against the stop task. For high-frequency devices (mice/trackballs at 125–1000 Hz), this produced billions of short-lived Task objects over days of uptime, fragmenting CPython's pymalloc arenas and causing monotonic RSS growth (~100 MB/day, observed 1.1 GB after 10 days). Replaced with a single `_stop_watcher` task that signals the shared `events_ready` Event, and `await events_ready.wait()` directly.

- **Autoload presets for all instances of the same device model**: When multiple pairing slots on a Logitech Unifying receiver expose the same device name (e.g. "Logitech MX Ergo", "Logitech MX Ergo 2"), only the first group matched the autoload config. `get_autoload_preset` now falls back to matching by the group's base device name, and `autoload()` iterates all discovered groups (filtering out input-remapper virtual devices) so every instance gets the preset.

## Test plan

- [x] Verify injection works for remapped devices after restart
- [ ] Monitor RSS of injector child process over hours/days — should stay flat
- [x] Test with multiple instances of the same device model (e.g. Unifying receiver with multiple slots)
- [ ] Test autoload with single-device config entry matching multiple group instances
- [ ] Verify input-remapper virtual devices are not passed to `_autoload`

### DIFF
--- a/inputremapper/configs/global_config.py
+++ b/inputremapper/configs/global_config.py
@@ -51,10 +51,21 @@ class GlobalConfig:
         """The folder containing this config."""
         return os.path.split(self.path)[0]
 
-    def get_autoload_preset(self, group_key: str) -> Optional[str]:
+    def get_autoload_preset(
+        self, group_key: str, group_name: Optional[str] = None
+    ) -> Optional[str]:
         # modifications are only allowed via the setter, because it needs to write
         # the config file too. Therefore return a copy to prevent inconsistencies.
-        return copy.deepcopy(self._config["autoload"].get(group_key))
+        #
+        # First try an exact match on the group key (e.g. "Logitech MX Ergo 2").
+        # If that fails and a group_name is provided, fall back to matching by the
+        # base device name (e.g. "Logitech MX Ergo"). This allows a single config
+        # entry to cover all instances of the same device model, such as multiple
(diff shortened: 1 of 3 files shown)


sezanzeb:
- As far as I can tell, this would require a new event to arrive before the stop_event would stop read_loop
- Test failing
- If two devices of the same make are connected, each one can have an individual preset to autoload. You'll have to navigate into the menu for the second device to enable autoloading there

~/.config/input-remapper-2/config.json:

```json
{
    "version": "2.2.0",
    "autoload": {
        "USB Optical Mouse 2": "autoload mouse 2",
        "USB Optical Mouse": "autoload mouse 1"
    }
}
```

Better create two PRs for this, if you can confirm that nothing breaks with this, and if the change still makes sense considering the point above.

sezanzeb:
Ah, sorry, I missed that you are setting events_ready once it is stopped via ensure_future.

Still, separate PR please.

And if possible, a script that reproduces the memory leak using a quickly iterating while loop would be nice, for me to reproduce this without having to keep my pc running for a few days.

Issue #1284, opened 2026-04-14
Bug: input-remapper-control causes USB device reset loop on devices with no profile (8BitDo Arcade Controller)


TheCoderGamer:
### Description

When a gamepad is connected that has no user-defined profile, `input-remapper-control` fails in a way that causes the USB device to reset. The device then reconnects, triggers udev again, `input-remapper-control` fails again, and the cycle repeats indefinitely every 3-4 seconds. The controller becomes completely unusable.

The expected behavior would be to fail silently and leave the device alone when no profile exists for it.

### Steps to reproduce

1. Have a gamepad connected with no input-remapper profile configured for it
2. Connect the device
3. Observe the device disconnecting and reconnecting in a loop

### Logs

udev triggering input-remapper-control, which fails and causes the reset loop:

```
(udev-worker)[12730]: input31: Process '/bin/input-remapper-control --command autoload --device ' failed with exit code 2.
(udev-worker)[14140]: js0: Process '/bin/input-remapper-control --command autoload --device /dev/input/js0' failed with exit code 5.
(udev-worker)[14123]: event28: Process '/bin/input-remapper-control --command autoload --device /dev/input/event28' failed with exit code 5.
```

The autoload desktop entry also attempts to access the device and fails:

```
input-remapper-gtk.desktop: Request to autoload for "8BitDo Ultimate 2 Wireless Controller"
gnome-shell: Could not open device /dev/input/event28: No such file or directory
```

The USB device cycling visible in dmesg (device number increments on every reconnect):

```
usb 1-5: New USB device found, idVendor=2dc8, idProduct=310b
usb 1-5: Product: 8BitDo Arcade Controller
hid-generic 0003:2DC8:310B.0033: hiddev99,hidraw5: USB HID v1.11 Device on usb-0000:01:00.0-5/input2
usb 1-5: USB disconnect, device number 68
usb 1-5: New USB device found, idVendor=2dc8, idProduct=310b   ← reconnect
...
```

### Additional context

The device (8BitDo Arcade Controller, idVendor=2dc8, idProduct=310b) has a firmware bug where it declares `bNumInterfaces 2` but numbers them 0 and 2 instead of 0 and 1. The kernel accepts it with a warning. This bug alone does not cause any issues — the device works fine without input-remapper. The reset loop only occurs when input-remapper is installed and running.

[...]

Issue #588, opened 2023-01-04
systemd-udevd services is getting failed with exit code 2 warning


tukanos:
I have not done anything to input-remapper-control but install it on by Debian Bookworm.

I'm getting the following warning when I list the systemd-udevd service.  First I want to ask, is that a normal behaviour?   If no, what can I do to get rid of the warning messages?  I tried to search for solution but did not find any.

I did run some debugging to see if I notice something obvious but did not.

The udev service warnings:

```
journalctl -u systemd-udevd.service -b
....
magnetron systemd-udevd[534]: input7: Process '/bin/input-remapper-control --command autoload --device ' failed with exit code 2.
Jan 04 10:24:31 magnetron systemd-udevd[541]: input0: Process '/bin/input-remapper-control --command autoload --device ' failed with exit code 2.
Jan 04 10:24:31 magnetron mtp-probe[639]: checking bus 1, device 7: "/sys/devices/pci0000:00/0000:00:14.0/usb1/1-8"
Jan 04 10:24:31 magnetron mtp-probe[639]: bus: 1, device: 7 was not an MTP device
Jan 04 10:24:31 magnetron systemd-udevd[549]: input14: Process '/bin/input-remapper-control --command autoload --device ' failed with exit code 2.
Jan 04 10:24:31 magnetron systemd-udevd[536]: input5: Process '/bin/input-remapper-control --command autoload --device ' failed with exit code 2.
Jan 04 10:24:31 magnetron systemd-udevd[558]: input19: Process '/bin/input-remapper-control --command autoload --device ' failed with exit code 2.
Jan 04 10:24:31 magnetron systemd-udevd[554]: input20: Process '/bin/input-remapper-control --command autoload --device ' failed with exit code 2.
Jan 04 10:24:31 magnetron systemd-udevd[539]: input13: Process '/bin/input-remapper-control --command autoload --device ' failed with exit code 2.
Jan 04 10:24:31 magnetron systemd-udevd[538]: input8: Process '/bin/input-remapper-control --command autoload --device ' failed with exit code 2.
Jan 04 10:24:31 magnetron systemd-udevd[548]: input10: Process '/bin/input-remapper-control --command autoload --device ' failed with exit code 2.
Jan 04 10:24:31 magnetron systemd-udevd[555]: input12: Process '/bin/input-remapper-control --command autoload --device ' failed with exit code 2.
[...]

PR #1285, opened 2026-04-18
Restrict 99-input-remapper.rules to event nodes

countgitmick:
## Summary

The udev rule fires synchronously for every add event in `SUBSYSTEM==input`. Most of those calls cannot succeed because the device is one of:

- a parent input class node with no `DEVNAME` (`input-remapper-control --command autoload --device ''` returns exit 2)
- a legacy `/dev/input/mouseN` or `/dev/input/mice` node (no group match, exit 1)
- a `/dev/input/jsN` joystick legacy node (no group match, exit 1)

Restricting the rule to `KERNEL=="event*"` keeps only the evdev nodes that input-remapper uses for group lookup. Every real input device exposes an event node, so no functionality is lost.

## Issues addressed

- **Fully addresses #588** (open since 2023-01-13, maintainer comment: "The udev rule is broken"). Eliminates all `input-remapper-control ... --device '' ... exit code 2` warnings that come from parent `inputN` class nodes firing the rule with empty `DEVNAME`.
- **Partially addresses #1284** (USB device reset loop on 8BitDo controller). The reporter's log shows three udev-rule firings per connect: `input31` exit 2, `js0` exit 5, and `event28` exit 5. This PR drops the first two but keeps `event28` firing. Whether the reset loop stops or persists depends on whether opening the `event*` node alone triggers the firmware bug on that device. The separate `/etc/xdg/autostart/input-remapper-autoload.desktop` autostart path that the reporter also disabled is outside the scope of this PR.

## Test plan

- [x] `udevadm control --reload-rules` then unplug/replug an input device
- [x] Verify `journalctl -u systemd-udevd` no longer shows `input-remapper-control [...] failed with exit code 1/2` from non-event nodes
- [x] Verify autoload still triggers for devices with a saved preset

### DIFF
--- a/data/99-input-remapper.rules
+++ b/data/99-input-remapper.rules
@@ -7,4 +7,4 @@
 #   journalctl -f
 # to get available variables:
 #   udevadm monitor --environment --udev --subsystem input
-ACTION=="add", SUBSYSTEM=="input", ENV{ID_PATH}!="platform-sound", RUN+="/bin/input-remapper-control --command autoload --device $env{DEVNAME}"
+ACTION=="add", SUBSYSTEM=="input", KERNEL=="event*", ENV{ID_PATH}!="platform-sound", RUN+="/bin/input-remapper-control --command autoload --device $env{DEVNAME}"


sezanzeb:
Thank you

Issue #1153, opened 2025-08-27
All mouse settings are seemingly overriden by the app to complete defaults once a mouse rebind is applied


gaygoblin:
As title states, rebinding any mouse input sets all mouse input settings from KDE to absolute defaults until it is unapplied. This does not *literally* change the settings, but temporarily sets them to the default values. Changing the settings while it is applied does not do anything.

wolfsilver00:
Happens to me too, remap any button and my dpi for example just jumps, the rest of my mapped buttons (from piper) stop working

TherinS:
Here is what I think you are saying:

When you rebind a mouse button using Input-Remapper and apply the setting, Input-Remapper overrides the custom functions you have set up in your mouse settings in the Operating System settings for the mouse, e.g. changing left-click button to right-click.  That is what I-R is supposed to do; rebind device keys to some output other than what the defaults for the OS are set to.

 I'm betting that is NOT what you are saying has happened, so perhaps be more descriptive so we can try to help.

gaygoblin:
@TherinS Nope! Not only was my mouse sensitivity entirely reset, but mouse-accel was also enabled. Any attempt to change those settings or revert them would be unsuccessful until I disabled the remap entirely. It entirely overrode my personal mouse configs. This doesn't feel like intended behavior - or, atleast, behavior that should be massively redone because changing a single input on my mouse shouldn't fuck with everything else about the mouse simultaneously.

TherinS:
@wolfsilver00 
It sounds like you are saying that after remapping some mouse buttons using Piper, you then map them again using Input-Remapper.  It is then that the previously mapped settings in Piper are overwritten by the mapping in Input-Remapper.  I suspect this is not what  you mean but that is what it looks like to me.  Please correct me and include the following information:

   1. the macro/string you are applying to the mouse button along with a description of what the macro is supposed to accomplish
    2. which mouse button it is being applied to, and
    3. the make and model of mouse being used.


@gaygoblin 
Here is what I think you are saying:
You have changed some system-wide mouse settings in the Operating System settings for the mouse, and among the settings changed is the DPI and mouse acceleration speed.  Did you have any button remapping in that settings menu?  If I am wrong, please correct me and include the following information:

   1. the macro/string you are applying to the mouse button along with a description of what the macro is supposed to accomplish
    2. which mouse button it is being applied to, and
    3. the make and model of mouse being used.

Splutty:
[...]

Issue #1048, opened 2025-01-31
mouse sensitivity resets when in use on kde plasma wayland (kubuntu)


the-nummer-zone:
whenever i have used input-remapper in x11 environments, it has worked without issue. but when i use it on kde on wayland, mouse sensitivity gets reset, causing the mouse to move much faster than usual. it does go back to normal once injecting stops.

not sure what other info to put here, so do ask if you need it.

sezanzeb:
This is an unfortunate and known issue for KDE and input-remapper:
- https://github.com/sezanzeb/input-remapper/issues/454
- https://github.com/sezanzeb/input-remapper/issues/358

NIICKTCHUNS:
I was having this issue today when I discovered Input Remapper, for my mouse I usually disable the Pointer acceleration over the settings, and the Input Remapper virtual mouse had this enabled because it is the default behaviour in KDE Plasma, disabling the Pointer acceleration solved the issue for me

![Image](https://github.com/user-attachments/assets/9096aa3e-7f1b-439e-ab5d-3d52fbf96666)

sezanzeb:
This has probably been fixed. It should reuse the speed of the original mouse now.

Issue #454, opened 2022-08-09
Starting injection resets mouse sensitivity.


DILeak:
1. `input-remapper 1.5.0 e31a1b2bc5d23fe13130afcc242063196335399f https://github.com/sezanzeb/input-remapper`
`python-evdev 1.4.0`
2. `Kubuntu 22.04`
3. `x11`
4. `Plasma`

When beginning injection on my mouse, including at startup, the sensitivity goes through the roof until I manually set the sensitivity in the mouse settings. It doesn't actually change the mouse setting itself in system settings, but it does make the system ignore it and use a higher sensitivity until it's changed again.

viscountexx:
Having the same issue on `NixOS  24.05`, `input-remapper 2.0.1`, `Wayland`, and `Plasma 5.27.11`. 

If it's helpful information, the sensitivity is defined in ~/.config/kcminputrc for a user, though I assume that just passes something along to libinput or something.

jonasBoss:
A workaround for this could be to create a mapping for the mouse X and Y axis to the mouse REL_X and REL_Y axis. Then you can change the sensitivity within Input-Remapper.
(See [Analog Axis](https://github.com/sezanzeb/input-remapper/blob/main/readme/usage.md#analog-axis))
![Bildschirmfoto_20241001_181804](https://github.com/user-attachments/assets/36573c0b-e33a-4eba-8ea4-839b376dc3e4)

sezanzeb:
This may have been fixed with the latest commit to main.  I'm also using KDE. Though I'm on wayland, not x11..

If not, please reopen this issue.

Issue #358, opened 2022-03-28
Sensitivity reset after injection


zerowhybr:
Hello,
I am facing an issue regarding pointer speed (mouse sensitivity) after injection happens.
I have set on KDE Plasma settings app to lower the sensitivity of my mouse, but after boot it does not persist as input-remapper injection ends and the pointer speed goes back to default. After I manually change the pointer speed on the settings to another value and back to what I want or click on stop injection it goes back to working normally.
Maybe add a way to control pointer speed on your app is a workaround. I don't know if it is possible to not touch pointer speed on injection.

Some logs:

1. `input-remapper-control --version`: 
```
input-remapper 1.4.2 5f53ae8c7b7e97dd02cf805cba6396637008c9e3 https://github.com/sezanzeb/input-remapper
python-evdev 1.4.0
```
2. which linux distro (ubuntu 20.04, manjaro, etc.): Arch Linux
3. `echo $XDG_SESSION_TYPE`: x11
4. which desktop environment (gnome, plasma, xfce4, etc.): KDE Plasma
5. `sudo ls -l /proc/1/exe`: `lrwxrwxrwx 1 root root 0 mar 28 10:24 /proc/1/exe -> /usr/lib/systemd/systemd`

6. `cat ~/.config/input-remapper/config.json`:
```
{
    "version": "1.4.2",
    "autoload": {
        "HID 1bcf:08a0 Mouse": "Default"
    },
    "macros": {
        "keystroke_sleep_ms": 10
    },
    "gamepad": {
        "joystick": {
            "non_linearity": 4,
            "pointer_speed": 80,
            "left_purpose": "none",
            "right_purpose": "none",
            "x_scroll_speed": 2,
            "y_scroll_speed": 0.5
        }
    }
}
```
7. `input-remapper-control --command hello`:
```
Connected to the service
Daemon answered with "hello"
Done
```
8. `systemctl status input-remapper -n 50`:
```
● input-remapper.service - Service to inject keycodes without the GUI application
     Loaded: loaded (/usr/lib/systemd/system/input-remapper.service; enabled; vendor preset: disabled)
     Active: active (running) since Mon 2022-03-28 10:24:35 -03; 17min ago
   Main PID: 509 (input-remapper-)
      Tasks: 5 (limit: 16695)
     Memory: 38.3M
        CPU: 12.990s
     CGroup: /system.slice/input-remapper.service
             ├─ 509 /usr/bin/python3 /usr/bin/input-remapper-service
[...]

PR #1290, opened 2026-05-09
fix: preserve forwarded mouse DPI metadata during injection

mizuikki:
## Summary

This fixes a pointer feel regression when injecting mouse devices.

Before this change, the forwarded virtual mouse created by input-remapper
could lose hwdb-derived properties such as `MOUSE_DPI`. On devices like the
Logitech G Pro, that caused the injected pointer to feel different from the
original device even though the event capabilities were copied.

## What changed

- keep the forwarded device name identical to the source device so system
  mouse hwdb rules can still match
- assign a dedicated `input-remapper/...` phys path to forwarded devices so
  they remain distinguishable without reusing the original hardware phys
- add a udev rule that restores `ID_BUS` for forwarded devices before
  `70-mouse.rules` runs, allowing `MOUSE_DPI` to be applied
- prevent autoload and device discovery from treating input-remapper-created
  devices as real hardware
- update tests and local uinput mocks to cover forwarded phys handling and
  input-remapper device filtering

## Why

On my system, enabling injection for a Logitech G Pro changed the effective
pointer curve/DPI feel. Inspecting the injected device showed that the
forwarded virtual mouse was missing the original device's `MOUSE_DPI`
metadata.

After this patch, the forwarded device keeps the original device name, gets
`ID_BUS=usb`, and successfully receives the same `MOUSE_DPI` value as the
physical mouse.

## Verification

Manual verification on Linux with a Logitech G Pro:

- before the patch, the forwarded device had no `MOUSE_DPI`
- after the patch, both the physical device and the forwarded device expose:

  `MOUSE_DPI=400@1000 *800@1000 1600@1000 3200@1000 6400@1000`

- pointer feel is now indistinguishable during injection in real use

## Notes

I also observed unrelated instability in the local test environment on
Python 3.14, but the DPI/metadata fix itself was validated on the real
device.

### DIFF
--- a/tests/lib/patches.py
+++ b/tests/lib/patches.py
@@ -210,13 +210,15 @@ uinputs = {}
 
 
 class UInputMock:
-    def __init__(self, events=None, name="unnamed", *args, **kwargs):
+    def __init__(
+        self, events=None, name="unnamed", phys="py-evdev-uinput", *args, **kwargs
+    ):
         self.fd = 0
         self.write_count = 0
         self.device = InputDevice("justdoit")
         self.name = name
         self.events = events
-        self.phys = kwargs.get("phys", "py-evdev-uinput")
+        self.phys = phys
         self.write_history = []
 
         global uinputs
--- a/tests/lib/patches.py
+++ b/tests/lib/patches.py
@@ -216,6 +216,7 @@ class UInputMock:
         self.device = InputDevice("justdoit")
         self.name = name
         self.events = events
+        self.phys = kwargs.get("phys", "py-evdev-uinput")
         self.write_history = []
 
         global uinputs
--- a/tests/unit/test_groups.py
+++ b/tests/unit/test_groups.py
@@ -176,7 +176,7 @@ class TestGroups(unittest.TestCase):
         }
 
         groups.refresh()
-        self.assertIsNone(groups.find(path="/foo/bar", include_inputremapper=True))
+        self.assertIsNone(groups.find(path="/foo/bar"))
 
     def test_is_inputremapper_device(self):
         device = evdev.InputDevice("/dev/input/event40")
--- a/tests/unit/test_groups.py
+++ b/tests/unit/test_groups.py
@@ -31,6 +31,7 @@ from inputremapper.groups import (
     classify,
     DeviceType,
     _Group,
+    is_inputremapper_device,
 )
 from tests.lib.fixtures import fixtures, keyboard_keys
 from tests.lib.test_setup import test_setup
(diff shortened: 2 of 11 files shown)


tests/lib/patches.py
```diff
@@ -216,6 +216,7 @@ def __init__(self, events=None, name="unnamed", *args, **kwargs):
         self.device = InputDevice("justdoit")
         self.name = name
         self.events = events
+        self.phys = kwargs.get("phys", "py-evdev-uinput")
         self.write_history = []
```
sezanzeb: I think phys could be added to the arguments, right?

`def __init__(self, events=None, name="unnamed", phys="py-evdev-uinput", *args, **kwargs):`

tests/unit/test_groups.py
```diff
@@ -167,6 +160,28 @@ def test_filter(self):
...
+                evdev.ecodes.EV_REL: [
+                    evdev.ecodes.REL_X,
+                    evdev.ecodes.REL_Y,
+                    evdev.ecodes.REL_WHEEL,
+                ],
+            },
+        }
+
+        groups.refresh()
+        self.assertIsNone(groups.find(path="/foo/bar", include_inputremapper=True))
```
sezanzeb: If `include_inputremapper` is True, shouldn't /foo/bar be included?

sezanzeb (changes requested):
Thank you!

There are a few old open issues about the mouse pointer speed, are they fixed with this?
- https://github.com/sezanzeb/input-remapper/issues/358
- https://github.com/sezanzeb/input-remapper/issues/454
- https://github.com/sezanzeb/input-remapper/issues/1048
- https://github.com/sezanzeb/input-remapper/issues/1153

mizuikki:
Thanks! I addressed the two inline comments in 89709c3:

- `UInputMock` now exposes `phys` as an explicit argument.
- The forwarded-device discovery test no longer uses `include_inputremapper=True`, since that option is about filtering already-discovered input-remapper groups. Devices with an `input-remapper` phys marker should be skipped during discovery.

All checks are passing on the latest commit.

About the linked pointer-speed issues: I would not claim this PR fixes all of them yet. This specifically fixes the case where the forwarded virtual mouse loses hwdb-derived metadata such as `MOUSE_DPI`, because the virtual device no longer matches the original mouse hwdb rules. That matches the issue I reproduced locally with my Logitech G Pro.

For reports caused by missing `MOUSE_DPI` on the forwarded device, this should help. Some of the linked reports also involve KDE/Wayland/X11 pointer speed, acceleration, or per-device settings not being applied to the forwarded virtual device, so those still need confirmation from affected users before closing the issues.

sezanzeb:
I think include_inputremapper can be removed from both filter and find

mizuikki:
Done, thanks. I removed `include_inputremapper` from both `filter` and `find` in 988e24c.

Since input-remapper-created devices are now skipped during discovery, the extra flag is no longer needed there. I also updated the reader client and group tests to use the simplified API.

5251c11 only fixes the resulting black formatting issue in `reader_client.py`. All checks are passing on the latest commit.

sezanzeb:
I just tried your changes locally, and mouse speed correctly stays the same (Kubuntu 25.10). I am happy about the code comments and that you also worked on the tests.

Just one more thing, probably: Now that filter doesn't filter anything anymore, lets rename it to get_groups. Then I think I'll merge this.

mizuikki:
Done, thanks. I renamed `filter()` to `get_groups()` in e82807b and updated the GUI callers and tests.

All checks are passing on the latest commit.

PR #1327, opened 2026-07-14
Bypass pkexec for reader-service if user has read access

shelterx:
For a smoother user experience, check if the current user has read permissions to '/dev/input/event*' devices (e.g. if they are in the 'input' group) and start the reader service directly without pkexec if so.

### DIFF
--- a/inputremapper/gui/reader_service.py
+++ b/inputremapper/gui/reader_service.py
@@ -39,6 +39,7 @@ different input-events into simple on/off events and sends them to the gui.
 from __future__ import annotations
 
 import asyncio
+import glob
 import logging
 import multiprocessing
 import os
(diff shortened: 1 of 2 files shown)


sezanzeb:
Thanks, it works, I'll merge this.

How about a unit test. I just quickly threw a draft together based on an existing test:

```py
    def test_no_pkexec_for_service(self):
        with patch.object(os, "access", lambda: True):
            os_system_mock = MagicMock(return_value=0)
            with patch.object(os, "system", os_system_mock):
                ReaderService.pkexec_reader_service()
                os_system_mock.assert_called_once_with("...")
```

into test_reader.py

Maybe using the newer syntax to use multiple `with` statements at once

shelterx:
@sezanzeb Well, it's beyond me but I pushed something that should work.

sezanzeb:
Looks good, tests are passing, thanks!

Were there difficulties when working with input-remappers test setup? Or other reasons?

shelterx:
All good, writing tests isn't something I usually do tho.

PR #1328, opened 2026-07-15
Add Global toggle support

shelterx:
Add a way to globally suspend and resume input remapper without losing its active configuration.

### DIFF
--- a/inputremapper/daemon.py
+++ b/inputremapper/daemon.py
@@ -121,7 +121,13 @@ class DaemonProxy(Protocol):  # pragma: no cover
 
     def start_injecting(self, group_key: str, preset: str) -> bool: ...
 
+    def get_running_preset(self, group_key: str) -> str: ...
+
     def stop_all(self) -> None: ...
+
+    def set_suspended(self, suspended: bool) -> None: ...
+
+    def is_suspended(self) -> bool: ...
 
     def set_config_dir(self, config_dir: str) -> None: ...
 
@@ -163,6 +169,16 @@ class Daemon:
                 </method>
                 <method name='stop_all'>
                 </method>
+                <method name='set_suspended'>
+                    <arg type='b' name='suspended' direction='in'/>
+                </method>
+                <method name='is_suspended'>
+                    <arg type='b' name='response' direction='out'/>
+                </method>
+                <method name='get_running_preset'>
+                    <arg type='s' name='group_key' direction='in'/>
+                    <arg type='s' name='response' direction='out'/>
+                </method>
                 <method name='set_config_dir'>
                     <arg type='s' name='config_dir' direction='in'/>
                 </method>
@@ -195,6 +211,8 @@ class Daemon:
         self.mapping_parser = mapping_parser
 
         self.injectors: Dict[str, Injector] = {}
+        self.suspended = False
+        self.suspended_presets: Dict[str, str] = {}
 
         self.config_dir = None
 
@@ -312,6 +330,9 @@ class Daemon:
 
     def stop_injecting(self, group_key: str) -> None:
         """Stop injecting the preset mappings for a single device."""
+        if group_key in self.suspended_presets:
+            del self.suspended_presets[group_key]
+
         if self.injectors.get(group_key) is None:
             logger.debug(
                 'Tried to stop injector, but none is running for group "%s"',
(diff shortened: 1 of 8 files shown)


inputremapper/injection/injector.py
```diff
@@ -211,8 +211,12 @@ def stop_injecting(self) -> None:
 
         Can be safely called from the main procss.
         """
-        logger.info('Stopping injecting keycodes for group "%s"', self.group.key)
```
sezanzeb, line 214: Is there a particular reason why this log before .send is removed?

inputremapper/daemon.py
```diff
@@ -195,6 +211,8 @@ def __init__(
         self.mapping_parser = mapping_parser
 
         self.injectors: Dict[str, Injector] = {}
+        self.suspended = False
+        self.should_run = {}
```
sezanzeb: Can this somehow be implemented using a queue called `suspended`?

I don't remember if an Injector object can be started after it was stopped. But if it can, you could use a python queue with these objects. If not, maybe a python queue with (group_key, preset) tuples.

Something like

```py
    def set_suspended(self, suspended: bool) -> None:
        ...
        for group_key, injector in self.injectors.items():
            self.suspended.push(injector)
            self.stop_injecting(group_key)
        ...
```

I'm not entirely sure what throws me off about should_run. Intuition or something.
- The name isn't very indicative of the feature that it serves (unlike a queue called "suspended")
- Having two datastructures that both contain running injections sounds like a risk for inconsistency bugs. For example, when stopping an injection, both datastructures need to be modified, and developers need to remember this, otherwise they cause bugs. However, a dict called "suspended" that is strictly contained within the `suspend_...` methods is easy to maintain/understand. During normal operation it is empty and can be forgotten about.
- It says "should_run", but until stuff is suspended it is actually running. Which makes it sound like should_run could be cleared of the contents since clearly stuff is running. should_run sounds a bit like a queue.

And keep it empty until stuff is actually suspended.

What do you think?

sezanzeb (commented):
Thanks. I'm assuming this is a feature you have been missing in input-remapper?

shelterx:
> Thanks. I'm assuming this is a feature you have been missing in input-remapper?

Yes, turning off all active remaps globally is something I've missed. I had this as a local change only but I thought I'd try and create a PR for it too.

One scenario could be, a user have remaps active when playing a game, then when finished or taking a break, instead of deactivating each remap individually or shutting down input remapper the user can just now just suspend it then easily resume it later.
I think it serves as a nice QoL improvement.

inputremapper/injection/injector.py
shelterx: Oh, this got lost when splitting up the changes into commits. It should be back again now.

inputremapper/daemon.py
shelterx: Indeed, renamed to suspended_presets, and it's now only populated at set_suspended(True) (snapshotted from self.injectors) and cleared at resume, empty the rest of the time.

shelterx:
Btw, I was very ambivalent regarding the naming of the UI setting. Do you think it's good or should I change it to something else?

inputremapper/daemon.py
```diff
@@ -327,6 +348,53 @@ def get_state(self, group_key: str) -> InjectorState:
...
+
+    def set_suspended(self, suspended: bool) -> None:
+        """Globally suspend or resume all remappings."""
+        if self.suspended == suspended:
+            return
+        self.suspended = suspended
+        if self.suspended:
+            logger.info("Suspending all active remappings")
+            for group_key, injector in list(self.injectors.items()):
+                # Only suspend injectors that are currently active (matching InjectorStateMessage.active())
```
sezanzeb: Last nitpick I think: This method is quite deep.

The contents of this loop could be extracted into a private method called `def _suspend_injection(self, group_key, injector)` (with the if negated into a return-early/guard-clause) (and with the appropriate type hints)

sezanzeb:
> Btw, I was very ambivalent regarding the naming of the UI setting. Do you think it's good or should I change it to something else?

I could imagine a toggle button that says "suspend", with the tooltip "press here to temporarily suspend/pause all active presets"

When clicked it is set to active and gets highlighted. No need to overwrite the buttons text in that case.

The placement is good.

shelterx:
> a toggle button that says "suspend", with the tooltip "press here to temporarily suspend/pause all active presets"

How about that the tooltip says "Pause all active presents" when active and dynamically change it "Enable all active presets" when inactive?

sezanzeb:
Ok

inputremapper/daemon.py
shelterx: Done.

sezanzeb:
Approved. Ready to merge?

shelterx:
Yes. However I can't do final functionality testing until tomorrow, in case we run into something unexpected.

inputremapper/gui/components/suspend_button.py
```diff
@@ -0,0 +1,106 @@
...
+            self._on_global_switch_toggled,
+        )
+
+        # Initialize the toggled state and tooltip
+        self._update_global_switch()
+
+    def _update_global_switch(self, *_args) -> None:
+        is_suspended = True
+        try:
+            is_suspended = self.controller.data_manager._daemon.is_suspended()
```
sezanzeb: This should still be wrapped, like set_suspended

shelterx: Should be ok now.

shelterx:
I did some final testing and it seems reliable.

sezanzeb:
Awesome, thank you!

Issue #1333, opened 2026-07-22
[Feature evaluation] Systray support


shelterx:
Just checking if there's any interest for optional systray support in input remapper, if the appindicator module isn't installed, systray won't be available.

Features:
- optional close to tray option
- tray access to suspend / resume
- maybe tray access to profiles?

cc: @sezanzeb

sezanzeb:
Sure, why not.

I have never looked into how tray icons work.
- If a new qt gui is ever written, will the tray icon continue to work, or does it need to be rewritten as well?
- Does this require using a toolkit (gtk or qt)?
- Does this keep the gui in memory?

shelterx:
1. Afaik the Systray UI part needs to be rewritten if Input Mapper goes QT but the core logic in the code is the same of course so some stuff can be reused. I haven't looked into QT at all since it's currently a GTK app.
2. Current implementation will use libayatana-appindicator since it provides best compatibility but it's optional, so if it isn't installed the systray won't be available. That's the idea anyway.
3. GUI isn't unloaded, I think adding such a feature adds complexity. It's possible it saves some memory when put to systray tho' but I haven't checked how much appindicator adds.

sezanzeb:
ok. As an example, the opensnitch system tray needs 290MB of ram and is part of the gui, even though I really only need it for the popups. That's a situation I'd like to avoid.

The input-remapper-gtk gui needs "only" 100MB (150MB when recording), but I have an aversion against wasted RAM. One the one hand being mad about RAM prices and on the other hand using electron to build apps, having a bunch of processes in background that many people don't need and making fancier (buggier) websites with more dependencies just doesn't seem to fit well together.

I'd imagine a system tray to potentially have these options:
- suspend, resume
- start gui (don't open, don't put to background, actually start. It might just need to run `input-remapper-gtk`, so there isn't much complexity. Closing it actually closes it). The system-tray is most likely a separate process then.
- Are nested context menus still a thing? Then there could be a list of mice, keyboards and gamepads (just the devices that were classified as these common types, to keep it short) and hovering them opens another list with presets.

shelterx:
Yeah, I want small memory footprints too. However it doesn't seem like having the systray in remapper gui itself adds much  memory at all, in fact it seems to reduce it with ~4Mb when the main app "goes to systray". Overall memory usage doesn't seem to increase very much, if at all.

1. Yes
[...]

PR #1336, opened 2026-07-25
Add Systray helper/support

shelterx:
Here's an attempt to implement optional systray support as discussed in PR #1333 
The current implementation is that systray can be ran standalone or it can be toggled to run from the GUI, the end results are the same, just different ways of doing it.

I tried to keep systray behavior how the users generally expect systray implementaions to behave.
The logic was the hardest part.

EDIT:
Regarding the ruff warnings, it seems like ruff was bumped from 0.15.22 -> 0.16.0.

### DIFF
--- a/bin/input-remapper-tray
+++ b/bin/input-remapper-tray
@@ -1,0 +1,26 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# input-remapper - GUI for device specific keyboard mappings
+# Copyright (C) 2026 sezanzeb <b8x45ygc9@mozmail.com>
+#
+# This file is part of input-remapper.
+#
+# input-remapper is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# input-remapper is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with input-remapper.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Starts the system tray helper."""
+
+from inputremapper.bin.input_remapper_tray import InputRemapperTrayBin
+
+if __name__ == "__main__":
+    InputRemapperTrayBin.main()
--- a/inputremapper/bin/input_remapper_tray.py
+++ b/inputremapper/bin/input_remapper_tray.py
@@ -54,6 +54,7 @@ except (ImportError, ValueError):
         pass
 from typing import Any, Dict, Optional, Tuple
 
+from inputremapper.bin.process_utils import ProcessUtils
 from inputremapper.configs.data import get_data_path
 from inputremapper.configs.global_config import GlobalConfig
 from inputremapper.configs.paths import PathUtils
@@ -90,8 +91,6 @@ class InputRemapperTrayBin:
         tray.run()
 
     def run(self) -> None:
-        from inputremapper.bin.process_utils import ProcessUtils
-
         if ProcessUtils.count_python_processes("input-remapper-tray") >= 2:
             logger.info(
                 "Another input-remapper tray helper is already running. Exiting."
@@ -384,8 +383,6 @@ class InputRemapperTrayBin:
             logger.setLevel(old_level)
 
     def _on_show_activate(self, _widget) -> None:
-        from inputremapper.bin.process_utils import ProcessUtils
-
         if ProcessUtils.count_python_processes("input-remapper-gtk") > 0:
             logger.info("input-remapper-gtk is already running, not spawning a new one")
             return
@@ -394,8 +391,6 @@ class InputRemapperTrayBin:
         subprocess.Popen(["input-remapper-gtk"])
 
     def _on_exit_activate(self, _widget) -> None:
-        from inputremapper.bin.process_utils import ProcessUtils
-
         terminated = ProcessUtils.terminate_python_processes("input-remapper-gtk")
         if terminated:
             logger.info("Terminated %d running GUI process(es)", terminated)
--- a/inputremapper/bin/input_remapper_tray.py
+++ b/inputremapper/bin/input_remapper_tray.py
@@ -1,0 +1,436 @@
+# -*- coding: utf-8 -*-
+# input-remapper - GUI for device specific keyboard mappings
+# Copyright (C) 2026 sezanzeb <b8x45ygc9@mozmail.com>
+#
+# This file is part of input-remapper.
+#
+# input-remapper is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your order) any later version.
+#
+# input-remapper is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with input-remapper.  If not, see <https://www.gnu.org/licenses/>.
+
+"""System tray helper process."""
+
+from __future__ import annotations
+
+import glob
+import os
+import subprocess
+import sys
+import threading
+
+import gi
+
+gi.require_version("Gtk", "3.0")
+gi.require_version("GLib", "2.0")
+from gi.repository import GLib, Gtk
+
+# Try importing AppIndicator
+HAS_APPINDICATOR = False
+AppIndicator: Any = None
+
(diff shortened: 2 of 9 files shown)


inputremapper/gui/user_interface.py
```diff
@@ -403,6 +411,39 @@ def on_gtk_shortcut(self, _, event: Gdk.EventKey):
                 pass
 
     def on_gtk_close(self, *_):
+        from inputremapper.gui.components.settings import HAS_APPINDICATOR
```
sezanzeb, line 414: Imports to the top of the file please

inputremapper/gui/user_interface.py
```diff
@@ -403,6 +411,39 @@ def on_gtk_shortcut(self, _, event: Gdk.EventKey):
...
+                    subprocess.Popen(["input-remapper-tray", "--gui-spawned"])
+            except Exception as e:
+                logger.error("Failed to spawn input-remapper-tray: %s", e)
+
+            self.controller.close()
+            return False
+
+        # If close to tray is disabled, terminate any running GUI-spawned tray helper process
+        try:
+            from inputremapper.bin.process_utils import ProcessUtils
```
sezanzeb, line 435: same

inputremapper/bin/process_utils.py
```diff
@@ -17,23 +17,41 @@
...
+from typing import List, Optional
 import psutil
 
 
 class ProcessUtils:
     @staticmethod
-    def count_python_processes(name: str) -> int:
+    def _find_matching_processes(
+        name: str, extra_match: Optional[str] = None
+    ) -> List[psutil.Process]:
```
sezanzeb, line 28: so it filters by name, and if extra_match is set, additionally by the second argument? extra_match confused me at first, so how about:

```py
    def _find_matching_processes(
        name: str,
        arguments: None | list[str] = None,
    ) -> List[psutil.Process]:
```

and

```py
if arguments is None or all(argument in cmdline[2:] for argument in arguments)
```

`continue` has somewhat of a bad reputation, but since the condition is getting more complex, it's worth it I think:

```py
                if len(cmdline) < 2:
                    continue

                if "python" not in cmdline[0]:
                    continue

                if name not in cmdline[1]:
                    continue

                if arguments and not all(argument in cmdline[2:] for argument in arguments):
                    continue

                matches.append(process)
```

inputremapper/bin/input_remapper_tray.py
```diff
@@ -0,0 +1,436 @@
...
+from inputremapper.logging.logger import logger
+
+
+class InputRemapperTrayBin:
+    def __init__(self, global_config: GlobalConfig):
+        self.global_config = global_config
+        self.daemon: Optional[DaemonProxy] = None
+        self.indicator: Any = None
+        self.is_supported = HAS_APPINDICATOR
+        self.toggle_item: Any = None
```
sezanzeb, line 73: No Any please

inputremapper/bin/input_remapper_tray.py
```diff
@@ -0,0 +1,436 @@
...
+from inputremapper.configs.paths import PathUtils
+from inputremapper.daemon import Daemon, DaemonProxy
+from inputremapper.groups import groups
+from inputremapper.gui.gettext import _
+from inputremapper.injection.injector import InjectorState
+from inputremapper.logging.logger import logger
+
+
+class InputRemapperTrayBin:
+    def __init__(self, global_config: GlobalConfig):
```
sezanzeb, line 68:
```suggestion
    def __init__(self, global_config: GlobalConfig) -> None:
```

inputremapper/bin/input_remapper_tray.py
```diff
@@ -0,0 +1,436 @@
...
+
+    @staticmethod
+    def main() -> None:
+        logger.update_verbosity(True)
+        global_config = GlobalConfig()
+        tray = InputRemapperTrayBin(global_config)
+        tray.run()
+
+    def run(self) -> None:
+        from inputremapper.bin.process_utils import ProcessUtils
```
sezanzeb, line 93: Import to the top

shelterx:
Memory usage comparison, it seems to be quite normal compared to a similar tray implementation:
```
   Process                        | VSZ (Virtual Memory)           | RSS (Physical RAM Used)
  --------------------------------|--------------------------------|----------------------------
   arch-update/lib/tray.py        | 430,016 KB (~430 MB)           | 80,844 KB (~80.8 MB)
   input-remapper-tray            | 679,700 KB (~680 MB)           | 79,744 KB (~79.7 MB)

```

sezanzeb:
I'll look at this soon, sorry for the delay

sezanzeb:
I just pushed tons of changes to fix all ruff issues to main, this should fix the pipeline.

sezanzeb (commented):
I started this review weeks ago but I think I didn't finish reviewing all the changes back then. I'll make another pass soon, once these are addressed.

shelterx:
Done. Rebased too, hope it went well... :P

inputremapper/gui/user_interface.py
```diff
@@ -402,6 +412,32 @@ def on_gtk_shortcut(self, _, event: Gdk.EventKey):
...
     def on_gtk_close(self, *_):
+        if (
+            HAS_APPINDICATOR
+            and self.controller.data_manager.global_config.get_systray()
+        ):
+            try:
+                if ProcessUtils.count_python_processes("input-remapper-tray") == 0:
+                    logger.info("Spawning detached system tray process")
+                    subprocess.Popen(["input-remapper-tray", "--gui-spawned"])
+            except Exception as e:
```
sezanzeb, line 423: Which exception exactly? https://docs.astral.sh/ruff/rules/blind-except/

(Also in the try/except below)

sezanzeb:
The tray process is logging

```
08:43:18.500734 7112 tray ERROR groups.py:493: Did not find any input device, possibly due to missing permissions
```

on my system. Are you in the input group?

My devices submenu is empty and says "No presets configured"

inputremapper/bin/input_remapper_tray.py
```diff
@@ -0,0 +1,433 @@
...
+            mtimes[""] = os.path.getmtime(base_dir)
+            for item in os.listdir(base_dir):
+                sub_path = os.path.join(base_dir, item)
+                if os.path.isdir(sub_path):
+                    mtimes[item] = os.path.getmtime(sub_path)
+        except Exception:  # noqa: S110
+            pass
+        return mtimes
+
+    def _refresh_groups_silently(self) -> None:
```
sezanzeb, line 376: Why is it important to silence discovery prints?

bin/input-remapper-tray
sezanzeb, line 1: How about a .desktop file to autostart the tray?

sezanzeb: How about a .desktop file to autostart the tray?

inputremapper/bin/input_remapper_tray.py
```diff
@@ -0,0 +1,433 @@
...
+        """Call groups.refresh() temporarily raising the log level to silence debug discovery prints."""
+        import logging
+
+        old_level = logger.level
+        logger.setLevel(logging.WARNING)
+        try:
+            groups.refresh()
```
sezanzeb, line 383: I think the deamon will need a method to scan devices and send it back to the tray. In the gui this happens via CMD_REFRESH_GROUPS and the ReaderService. The gui should then also be changed to use the dbus method of the daemon for this.

shelterx:
> Are you in the input group?

Yeah, I'm in the input group... and /etc/udev/rules.d/60-uinput.rules sets:
KERNEL=="uinput", GROUP="input", MODE="0660"

It was early on I messed around with input permissions, but I think I added the above myself.
`error: No package owns /etc/udev/rules.d/60-uinput.rules`

inputremapper/bin/input_remapper_tray.py
shelterx: It's to reduce log spam, not sure if you want it visible during normal operartion.

shelterx: logger.update_verbosity is true for systray, what we can do is hide it behind a debug flag?
Though with the dbus change that log will be moved.

sezanzeb: I think _poll_state would only be important as long as the menu is open, right? If yes, this would also reduce log spam.

shelterx: I'll hold this open for now, I want dbus the change to land first.

inputremapper/bin/input_remapper_tray.py
shelterx: Gave it an attempt. Also added a signal to sync the suspend button in the GUI with the tray state.
One thing I noticed tho', if it's suspended it doesn't show which mappings are "active", it unchecks them in the tray,. Then when you resume the injection it re-checks them... might be confusing to the user but it seems to be related to the suspend feature.

bin/input-remapper-tray
shelterx: What will the user expect? 
If we just add it as is with a shortcut, it will always start the tray, however it will disappear when the user closes the GUI IF close to systray is disabled.
Perhaps we should sync that "state" instead, if close to systray is disabled it should never fire up the systray.
And rename it to "Enable systray" instead?

PR #1341, opened 2026-08-03
Fix joystick for some gamepads

sezanzeb:
On my new kubuntu 26.04  installation the xbox one gamepad seems to work out of the box over bluetooth. However, I couldn't record stuff with the joystick.

If I remember correctly, without xpadneo I couldn't even connect the gamepad. I'm assuming a newer kernel fixed this, and that I get different capabilities for the joystick now, that trigger a bug possibly introduced in https://github.com/sezanzeb/input-remapper/pull/1274


sezanzeb:
- [x] Write test that reproduces the issue. `if abs(event.value - mid_point) < threshold:` broke, because `threshold` was an absolute value (Meaning it included the mid-point, and was therefore a very large value), whereas event.value - mid_point is a small delta value. It then always set `pressed` to `False`. It therefore, after #1274, only worked for gamepads that have a -min to +max range, but not for ones with a 0 to +max range.

PR #1349, opened 2026-08-23
add ExecStop to service

Renari:
Package upgrades can cause a race condition when restarting `input-remapper` before the previous daemon releases its D-Bus name. This adds `ExecStop` using the daemon’s `quit` command to ensure it is released and treats the expected `NoReply` as success.

This issue would cause `arch-update` to report a failure restarting the `input-remapper` service every time I update.

### DIFF
--- a/inputremapper/bin/input_remapper_control.py
+++ b/inputremapper/bin/input_remapper_control.py
@@ -26,10 +26,7 @@ import subprocess
 import sys
 from enum import Enum
 
-import gi
-
-gi.require_version("GLib", "2.0")
-from gi.repository import GLib
+from dasbus.error import DBusError
 
 from inputremapper.configs.global_config import GlobalConfig
 from inputremapper.configs.migrations import Migrations
@@ -232,8 +229,8 @@ class InputRemapperControlBin:
     def _quit(self) -> None:
         try:
             self.daemon.quit()
-        except GLib.GError as error:
-            if "NoReply" in str(error):
+        except DBusError as error:
+            if error.dbus_name == "org.freedesktop.DBus.Error.NoReply":
                 # The daemon is expected to terminate, so there won't be a reply.
                 return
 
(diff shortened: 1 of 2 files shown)


inputremapper/bin/input_remapper_control.py
```diff
@@ -233,8 +234,10 @@ def _stop(self, device: str) -> None:
...
         try:
             self.daemon.quit()
-        except GLib.GError as error:
-            if "NoReply" in str(error):
+        except (DBusError, GLib.GError) as error:
+            if getattr(
+                error, "dbus_name", None
+            ) == "org.freedesktop.DBus.Error.NoReply" or "NoReply" in str(error):
                 # The daemon is expected to terminate, so there won't be a reply.
                 return
```
sezanzeb: If you use multiple except for the two error classes, then it should be possible to do this without getattr.

I generally avoid getattr and the likes as much as possible, because they make navigating code via symbols or search queries like `.dbus_name` impossible, and mypy cannot check getattr for errors.

sezanzeb (changes requested):
Thank you. I'll merge this as soon as the comment is addressed.

sezanzeb:
I fixed the lint pipeline on main yesterday, please merge main into your branch.

Renari:
Sorry when, when I rebased and pushed, it pushed the commit from your main instead of my changes as well, thus closing the PR due to no changes. Here's the change you requested.

sezanzeb:
I just realized the GError was a leftover from pydbus when I switched to dasbus. It's a regression of https://github.com/sezanzeb/input-remapper/pull/1201

Using DBusError here like you propose is correct, as a replacement of the Glib.GError. Sorry for the confusion, please remove the deprecated handling of the GError completely, then I'll merge.

Renari:
Done

sezanzeb:
Thank you!

PR #1351, opened 2026-08-26
Fix hold_keys dropping duplicate keys in + syntax (#1347)

### LINKED ISSUE #1347: KEY_3+KEY_3 outputs 3. not 33. No double numbers allowed? 😳
I want a button for a long number. Works fine as long as there are no double numbers. Am I overlooking something?

desired output:
4355941285

input:
KEY_4+KEY_3+KEY_5+KEY_5+KEY_9+KEY_4+KEY_1+KEY_2+KEY_8+KEY_5

actual output:
4359128

Notice only the first 5 and first 4 are in the output.

So numbers can only be used once?

3+KEY_3 also gives one 3. Adding space or deviders also does not make it work.

How to get output 43559412805?

arturomaciaas:
Fixes #1347 

**Problem**
When using the + syntax with repeated keys (e.g., KEY_4+KEY_3+KEY_5+KEY_5+KEY_9), duplicate keys were silently dropped because pressing an already-held key is a no-op at the kernel level.

**Solution**
The fix releases a key before re-pressing it when a duplicate is encountered in hold_keys, so that sequences like KEY_5+KEY_5 properly output both keystrokes.

**What was changed**
- inputremapper/injection/macros/tasks/hold_keys.py — track held codes and release before re-pressing duplicates; release each unique code once on trigger release
- tests/unit/test_macros/test_hold_keys.py — added tests for duplicate keys in both hold_keys() and + syntax

Before: KEY_4+KEY_3+KEY_5+KEY_5+KEY_9+KEY_4+KEY_1+KEY_2+KEY_8+KEY_5 = 4359128
After: KEY_4+KEY_3+KEY_5+KEY_5+KEY_9+KEY_4+KEY_1+KEY_2+KEY_8+KEY_5 = 4355941285

### DIFF
--- a/inputremapper/injection/macros/tasks/hold_keys.py
+++ b/inputremapper/injection/macros/tasks/hold_keys.py
@@ -44,12 +44,26 @@ class HoldKeysTask(Task):
 
         codes = [keyboard_layout.get(symbol) for symbol in symbols]
 
+        held_codes = set()
         for code in codes:
+            if code in held_codes:
+                # Release and re-press to support duplicate keys in a sequence
+                callback(EV_KEY, code, 0)
+                await self.keycode_pause()
             callback(EV_KEY, code, 1)
+            held_codes.add(code)
             await self.keycode_pause()
 
         await self._trigger_release_event.wait()
 
-        for code in codes[::-1]:
+        # Release each unique code once, in reverse order of first appearance
+        seen = set()
+        unique_codes_reversed = []
+        for code in reversed(codes):
+            if code not in seen:
+                seen.add(code)
+                unique_codes_reversed.append(code)
+
+        for code in unique_codes_reversed:
             callback(EV_KEY, code, 0)
             await self.keycode_pause()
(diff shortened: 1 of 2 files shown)


sezanzeb:
Looking at https://github.com/sezanzeb/input-remapper/issues/1347, what about a macro like

- `write(43559412805)` (which would be impossible to use with symbols like KEY_A)
- Or `keys(4, 3, 5, 5, 9, 4, 1, 2, 8, 0, 5)`, or maybe `sequence(4, 3, 5, ...)`, or `write(4, 3, 5, ...)`

This would be easier to use than the only existing solution, which is `key(4).key(3).key(5)...`

I'm not sure if releasing a key, even if it is duplicate, contradicts the intention of hold_keys. Maybe this should be an error case instead. A new macro function that does this explicitly might be less surprising for users.

sezanzeb:
or a new special syntax like

`4 3 5 5 9 KEY_A 4 1 2 KEY_ESC 8 0 5`

or

`4, 3, 5, 5, 9, KEY_A, 4, 1, 2, KEY_ESC, 8, 0, 5`

this could maybe be combined with `+`:

`a b c + a b d` / `a,b,c + a,b,d` would combine a, b and c, release them, then combine a, b and d

PR #1352, opened 2026-08-27
Add while_eq macro: variable-driven loop stoppable from another mapping

timcharper:
## Problem

I wanted an autoclicker toggled on by one key and stopped by a *different*
key (e.g. Shift+F4 starts clicking, F4 stops it and also releases a
held button from a third mapping). `toggle()` doesn't fit: it only
starts/stops on a second press of the *same* trigger, so an unrelated
key has no way to reach it.

The natural workaround is `repeat(N, if_eq($var, 1, click, wait))` driven
by a shared `set()`/`if_eq()` variable — but this runs into a sharper
problem than just "clicking never stops immediately". `RepeatTask.run()`
reads its count once and loops a plain `for _ in range(repeats)`, with no
way to exit early:

https://github.com/sezanzeb/input-remapper/blob/2.2.1/inputremapper/injection/macros/tasks/repeat.py#L44-L49

Meanwhile `Macro.run()` refuses to start a second execution while the
first is still "running":

https://github.com/sezanzeb/input-remapper/blob/2.2.1/inputremapper/injection/macros/macro.py#L83-L87

So even after the shared variable flips to stop the clicking, the macro
instance keeps occupying that mapping's "running" slot until the fixed
repeat count exhausts — which can be minutes depending on how the count
was sized — silently swallowing every re-press of the trigger key in the
meantime. From the user's perspective: works once, then appears to just
stop responding.

## Fix

`while_eq(variable, value, macro)` re-reads the variable every iteration
(like `if_eq` does) instead of deciding a count up front (like `repeat`
does), so `run()` actually *returns* as soon as the condition no longer
holds:

```ts
set(clicking, 1).while_eq(clicking, 1, key(BTN_LEFT).wait(50))
# a different mapping: set(clicking, 0) stops it, and the macro is
# immediately retriggerable — it isn't still "running" a leftover budget
```

This required a genuinely new task rather than composing existing ones —
none of `repeat`/`hold`/`toggle`/`if_tap`/`if_single` support an
externally-driven, unbounded, promptly-exiting loop; `hold`/`toggle`'s
early-exit mechanism (`press_trigger()`/`release_trigger()`) is scoped to
the *same* trigger that started them, and `repeat`'s bound is fixed at
start with no way to shorten it once running.

## Changes

- `inputremapper/injection/macros/tasks/while_eq.py` — new task
- `inputremapper/injection/macros/parse.py` — registers `while_eq` in `TASK_CLASSES`
- `readme/macros.md` — docs entry alongside `toggle`
- `tests/unit/test_macros/test_while_eq.py` — covers: no-op when the
  condition starts false, stopping promptly when the variable is changed
  by other code mid-loop, and immediate re-triggerability after stopping

Ran the full `tests/unit/test_macros` suite locally (134 tests, all
passing) plus the new file on its own.

Happy to adjust naming/placement/docs wording if you'd prefer something
different — `while_eq` was chosen to mirror `if_eq`'s naming rather than
introduce an unrelated verb (I'd originally called it `until`, which is
backwards — `until` conventionally means "loop while false", and this
loops while *true*).

### DIFF
--- a/inputremapper/injection/macros/parse.py
+++ b/inputremapper/injection/macros/parse.py
@@ -49,6 +49,7 @@ from inputremapper.injection.macros.tasks.mouse_xy import MouseXYTask
 from inputremapper.injection.macros.tasks.repeat import RepeatTask
 from inputremapper.injection.macros.tasks.set import SetTask
 from inputremapper.injection.macros.tasks.toggle import ToggleTask
+from inputremapper.injection.macros.tasks.while_eq import WhileEqTask
 from inputremapper.injection.macros.tasks.wait import WaitTask
 from inputremapper.injection.macros.tasks.wheel import WheelTask
 from inputremapper.logging.logger import logger
@@ -63,6 +64,7 @@ class Parser:
         "modify": ModifyTask,
         "repeat": RepeatTask,
         "toggle": ToggleTask,
+        "while_eq": WhileEqTask,
         "key": KeyTask,
         "key_down": KeyDownTask,
         "key_up": KeyUpTask,
--- a/inputremapper/injection/macros/tasks/while_eq.py
+++ b/inputremapper/injection/macros/tasks/while_eq.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# input-remapper - GUI for device specific keyboard mappings
+# Copyright (C) 2025 sezanzeb <b8x45ygc9@mozmail.com>
+#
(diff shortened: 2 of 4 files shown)


sezanzeb:
Thanks! Your implementation of while_eq mirrors the deprecated ifeq macro (`variable` and `value` arguments), and instead should probably be changed to mirror if_eq (`value_1`, `value_2`).

I fixed the lint pipeline yesterday, just merge main into your branch

